PR #1, opened 2019-02-07
Add nfs csi driver

prateekpandey14:


pkg/nfs/driver.go
```diff
@@ -0,0 +1,78 @@
...
+limitations under the License.
+*/
+
+package nfs
+
+import (
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/golang/glog"
+
+	csicommon "github.com/kubernetes-csi/drivers/pkg/csi-common"
```
mathu97: I believe "github.com/kubernetes-csi/drivers" is supposed to be deprecated once all the drivers have been moved to their own respective repo. Maybe we should not use this library?

msau42:
Is it possible to move this over and retain the commit history?

prateekpandey14:
@msau42 I will retain the commit history.. thanks

pkg/nfs/driver.go
prateekpandey14: @mathu97 right , but i can see there is `https://github.com/kubernetes-csi/csi-lib-common` where we will move out the csi-common code base , which can be used later by the other drivers.

mathu97: Oh okay, i didn't realize that there was such a repo. Sounds good.

msau42:
cc @bswartz

prateekpandey14:
Raised PR https://github.com/kubernetes-csi/csi-driver-nfs/pull/2 with retain commit history

PR #2, opened 2019-02-12
Moves nfs csi-driver from kubernetes-csi/drivers mono-repo

prateekpandey14:
This PR is to setup and migrate the NFS CSI driver from  [drivers](https://github.com/kubernetes-csi/drivers)  mono repo to its own repository.

* Copies source from [drivers](https://github.com/kubernetes-csi/drivers) mono repo
* Rearrange packages and names
* Adjust Makefile and other build files
* Update the deploy files with required images
* Update  README documentation


pkg/nfs/deploy/kubernetes/csi-attacher-nfsplugin.yaml
```diff
@@ -0,0 +1,64 @@
+# This YAML file contains attacher & csi driver API objects that are necessary
```
saad-ali: These files look like they are a duplicate of `deploy/kubernetes/*`

app/nfsplugin/main.go
```diff
@@ -0,0 +1,70 @@
+/*
```
saad-ali: We can remove `nfsplugin` since this repo will only contain the nfs driver.

pkg/nfs/deploy/kubernetes/csi-attacher-nfsplugin.yaml
prateekpandey14: ahh .. missed to delete them. will fix it.

app/nfsplugin/main.go
prateekpandey14: Same thing here. thanks

pkg/nfs/deploy/kubernetes/csi-attacher-nfsplugin.yaml
prateekpandey14: Done !

app/nfsplugin/main.go
prateekpandey14: Done !

PR #10, opened 2019-04-15
add Kubernetes End to End tests to the NFS Driver

mathu97:
**What type of PR is this?**

/kind feature

This PR adds End to End tests to the NFS plugin. This is done by importing in-tree Kubernetes CSI storage E2E tests. The NFS CSI plugin could be an example for other CSI plugin authors who also want to leverage the in-tree storage tests for their plugins.

This would be a stable example for the [CSI e2e testing documentation](https://github.com/kubernetes-csi/docs/pull/143#discussion_r274652302)

- This PR updates the driver from Kubernetes-1.12 to Kubernetes-1.14

- Users would be able to run End To End tests on the driver by using the command: `go test -v ./cmd/nfsplugin/ -ginkgo.v -ginkgo.progress --kubeconfig=/var/run/kubernetes/admin.kubeconfig -timeout=0`

```release-note

```


mathu97:
@wongma7 @msau42 please take a look.

Gopkg.toml
```diff
@@ -68,10 +68,18 @@
   name = "github.com/json-iterator/go"
   version = "1.1.4"
 
+[[override]]
+  name = "k8s.io/client-go"
+  branch = "master"
```
msau42: do we need master or can kubernetes-1.14.0 work?

Gopkg.toml
```diff
@@ -68,10 +68,18 @@
...
+
 [[override]]
   name = "gopkg.in/square/go-jose.v2"
   version = "2.1.7"
 
+[[override]]
+  source = "https://github.com/fsnotify/fsnotify/archive/v1.4.7.tar.gz"
```
msau42: is this actually needed?

test/csi-volumes.go
```diff
@@ -0,0 +1,43 @@
...
+	"k8s.io/kubernetes/test/e2e/storage/utils"
+	"path"
+)
+
+var CSITestSuites = []func() testsuites.TestSuite{
+	testsuites.InitVolumesTestSuite,
+	testsuites.InitVolumeIOTestSuite,
+	testsuites.InitVolumeModeTestSuite,
+	testsuites.InitSubPathTestSuite,
+	testsuites.InitProvisioningTestSuite,
```
msau42: Would it be useful to define a global variable in the test framework that contains all the test suites? That way drivers don't need to manually copy them?

test/nfs-testdriver.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (n *nfsDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTestConfig, func()) {
+	config := &testsuites.PerTestConfig{
+		Driver:    n,
+		Prefix:    "nfs",
+		Framework: f,
+	}
+
+	//Install the nfs driver from the manifests
+	cleanup, err := config.Framework.CreateFromManifests(nil, n.manifests...)
```
msau42: Instead of creating a new nfs driver for every test case, can we bring up the nfs driver once in the BeforeSuite()?

Gopkg.toml
```diff
@@ -42,22 +42,22 @@
 
 [[constraint]]
   name = "k8s.io/kubernetes"
-  version = "v1.12.0"
+  version = "v1.14.0"
```
msau42: Can you put the vendor updates into a separate commit?

pkg/nfs/controllerserver.go
```diff
@@ -0,0 +1,23 @@
...
+	*csicommon.DefaultControllerServer
+}
+
+func (cs ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi.ControllerExpandVolumeRequest) (*csi.ControllerExpandVolumeResponse, error) {
+	return nil, status.Error(codes.Unimplemented, "")
+}
+
+func getControllerServer(csiDriver *csicommon.CSIDriver) ControllerServer {
+	return ControllerServer{
+		csicommon.NewDefaultControllerServer(csiDriver),
```
msau42: I don't think we've updated csicommon for csi spec 1.1 (and we don't intend to. We want to deprecate csicommon).

README.md
```diff
@@ -66,6 +66,13 @@ $ csc node get-id --endpoint tcp://127.0.0.1:10000
 CSINode
 ```
 
+## Running Kubernetes End To End tests on the NFS Driver
+1) Stand up a local cluster `ALLOW_PRIVILEGED=1 hack/local-up-cluster.sh`
+2) Build the nfs driver by running `make`
+3) Create NFS Driver Image, where the image tag would be whatever that is required by your YAML deployment files        `docker build -t quay.io/k8scsi/nfsplugin:v1.0.0 .`
+4) Run E2E Tests using the following command: `go test -v ./cmd/nfsplugin/ -ginkgo.v -ginkgo.progress --kubeconfig=/var/run/kubernetes/admin.kubeconfig -timeout=0`
```
msau42: How does someone run the e2e tests with their own private images?

Gopkg.toml
mathu97: I was having an issue similar to [this](https://github.com/kubernetes/kubernetes/issues/76304), but using master solved it. I'll look more into it to see if we can avoid using master.

Gopkg.toml
mathu97: when using dep ensure, I was getting a `panic: version queue is empty` error, and adding this solved it. Reference: https://github.com/golang/dep/issues/1799

test/csi-volumes.go
mathu97: Yes I think that would be useful. Would there be any case where a driver would not want to run some of these test suites?

pkg/nfs/controllerserver.go
mathu97: Are we still trying to move csicommon to [kubernetes-csi/csi-lib-common](https://github.com/kubernetes-csi/csi-lib-common)?

msau42: No we're just going to deprecate it. It doesn't seem to provide much value.

test/csi-volumes.go
msau42: It shouldn't need to. The test cases are supposed to check for driver capabilities and only run the tests that the driver can support.

pkg/nfs/controllerserver.go
mathu97: okay, also should we stick to csi spec 1.0?

msau42: 1.1 is fine too. The resizing stuff will be unimplemented

README.md
mathu97: you would need to add the required YAML files for that private image in [deploy/kubernetes](https://github.com/kubernetes-csi/csi-driver-nfs/tree/master/deploy/kubernetes), as those are the files that the e2e test uses to deploy the driver. I'll add that to the documentation.

test/nfs-testdriver.go
mathu97: Currently the testdriver automatically creates a new namespace with a randomly generated name for each testcase, and I think the driver needs to be in that same namespace for it to be accessible by the tests. This would be hard to do in the BeforeSuite() if we don't know what the randomly generated namespace is called.

msau42: I think if you don't use CreateFromManifests, you can just bring up the driver directly in some default namespace.  This is how we configure pd csi driver to run e2e externally and not bring up the driver for every test:

https://github.com/kubernetes/kubernetes/blob/51db0fe21a816b0818df571ad70386864d83c722/test/e2e/storage/drivers/csi.go#L444

mathu97: Ya that makes sense. So we would bring up the driver in a default namespace, instead of creating a new namespace for each test case right?

msau42: Yeah the driver could be brought up in the default namespace.  Test case application pods will still come up in their own namespaces though

mathu97: Is there something wrong with bringing up the Test case application pods in the same default namespace as the driver? I'm trying to run them in separate namespaces (driver in default, and test case application pods in its own), but the test cases either timeout or take a long time to finish.

msau42: In a real production cluster, the driver is generally installed in its own namespace and user applications will be in their own namespaces too.  So cross namespace access must be possible.  If you are seeing issues I would try to dig in more. It could indicate that the test is dependent on the driver deployment more than it needs to

mathu97: @msau42 I'm still testing it, but it seems to be working fine. Would it be better to have the user manually deploy the driver in a default namespace, instead of having the tests use these specific [YAML files](https://github.com/kubernetes-csi/csi-driver-nfs/pull/10/commits/892c7a866e88708d11f0d0902d3c45787c101277#diff-c69d8975ffcf100ed7be43c0052d993bR59) to do it? This way I think it would be easier for users to run e2e tests with their own private images.

wongma7: there are 2 situations, 1 is CI which is when one would want to redeploy `latest` yamls before every test, 2 is testing an already-installed driver. It would be nice if this example could demonstrate both? Is there a way for the test to detect if the driver is already installed?

msau42: Even in a CI situation, would you want to deploy the yamls every time?  Nothing about the driver yamls should have changed in between each run?

wongma7: for now auto-deploying whatever is in the yaml folder every BeforeSuite is fine. If the user wants to test a different image version they can edit the yamls. In the future maybe there will be a flag for whether  the test should deploy the driver or not if it already exists

wongma7: @msau42 it's possible, like e.g. hostpath driver had some hardcoded paths that were changed so the yaml changed

wongma7: oh, I misspoke didn't mean before every test i mean before every testsuite. We agree on that already

Gopkg.toml
mathu97: done

test/csi-volumes.go
msau42: We also have Snapshot and MultiVolume test suites, although they don't seem applicable to this nfs driver. Should be include them anyway and comment them out?

test/nfs-testdriver.go
```diff
@@ -0,0 +1,154 @@
...
+var _ testsuites.PreprovisionedVolumeTestDriver = &nfsDriver{}
+var _ testsuites.PreprovisionedPVTestDriver = &nfsDriver{}
+
+func (n *nfsDriver) GetDriverInfo() *testsuites.DriverInfo {
+	return &n.driverInfo
+}
+
+func (n *nfsDriver) SkipUnsupportedTest(pattern testpatterns.TestPattern) {
+	if pattern.VolType == testpatterns.DynamicPV {
+		framework.Skipf("NFS Driver does not support dynamic provisioning -- skipping")
```
msau42: in the future it may be cool to add something like the nfs-client provisioner to the sample driver, and then we can unlock these test cases as well

test/nfs-testdriver.go
msau42: It looks like this is still bringing up the nfs driver for every test case?

README.md
msau42: I don't see the updated documentation pointing to which yamls need to be changed

test/csi-volumes.go
mathu97: Ya sure, since this would be used as an example right. Also, I think MultiVolume testsuite is not in 1.14?

mathu97: I guess it doesn't matter since it'll still be commented out.

test/nfs-testdriver.go
mathu97: Sorry, I still haven't pushed it.

README.md
hoyho: should we build local image on each node? since the image is not publish to quay.io, if the plugin will run on each node, then should build or import the image to all nodes?

test/nfs-testdriver.go
mathu97: +1

README.md
mathu97: just pushed the updated docs

mathu97:
@msau42 let me know what you think about the new changes

msau42:
Sorry for the delay.  Thanks for working on this!

/lgtm
/approve

PR #14, opened 2019-06-07
WIP: Prow testing

pohly:
**What type of PR is this?**
/kind feature

With `.prow.sh` in place, Prow jobs for this driver can be enabled in https://github.com/kubernetes/test-infra/blob/d04b617e2f4d4995dc7f650da8b720d272910855/config/jobs/kubernetes-csi/gen-jobs.sh#L26-L33

Then we will have pre-submit checks as well as testing on different Kubernetes releases.

Because it is expected by `prow.sh`, a deployment script derived from the one in the hostpath driver repo also gets added, which may be useful also for users.

```release-note
more flexible deployment via script
```


pohly:
This is WIP because `release-tools/prow.sh` had to be modified and not all tests passed for me. If this is useful, that `prow.sh` change needs to go into `csi-release-tools` first.

To try out this PR, check out the branch into an *empty* GOPATH, then run `.prow.sh`. It will leave files behind in that GOPATH as well as check out and modify other repos. Also, the kind cluster is kept running. It needs to be stopped manually with `kind delete cluster --name csi-prow`.

pohly:
@mathu97 can you check why some of the tests are failing? See my comment above for how to run them yourself.

mathu97:
> @mathu97 can you check why some of the tests are failing? See my comment above for how to run them yourself.

Trying it out

mathu97:
@pohly all the pre-provisioned tests are timing out, when I tried to run it.

.prow.sh
```diff
@@ -0,0 +1,38 @@
...
+
+. release-tools/prow.sh
+
+# Install custom E2E test suite as bin/tests.
+install_e2e () {
+    if [ -e "${CSI_PROW_WORK}/e2e.test" ]; then
+        return
+    fi
+
+    make builds-tests
```
mathu97: typo, should be `make build-tests`

deploy/kubernetes/deploy.sh
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env bash
+
+# This script captures the steps required to successfully
+# deploy the hostpath plugin driver.  This should be considered
```
mathu97: typo: hostpath -> NFS

pohly:
> all the pre-provisioned tests are timing out, when I tried to run it.

Yes, I think that's also what I was seeing. But why are they timing out?

.prow.sh
pohly: Huh, how the heck did that work?

Ah, missing error checking combined with building `bin/tests` earlier. Fixed and force-pushed.

pohly:
@mathu97 I added another commit that simplifies `cmd/tests/nfs-e2e.go`. I think it is easier to understand that way. I can also submit that separately, if you want.

mathu97:
> > all the pre-provisioned tests are timing out, when I tried to run it.
> 
> Yes, I think that's also what I was seeing. But why are they timing out?

Looking into it

cmd/tests/nfs-e2e.go
```diff
@@ -14,25 +14,16 @@ limitations under the License.
...
-func Test(t *testing.T) {
-	flag.Parse()
-	RegisterFailHandler(Fail)
-	RunSpecs(t, "CSI Suite")
-}
-
-func main() {
-	Test(&testing.T{})
+	gomega.RegisterFailHandler(ginkgo.Fail)
+	ginkgo.RunSpecs(&testing.T{}, "CSI Suite")
```
mathu97: ya this makes more sense!

README.md
```diff
@@ -74,18 +75,67 @@ For Fedora/RHEL clusters, the following might be required:
...
-3) Install the Driver: `kubectl create -f deploy/kubernetes`
-4) Build E2E test binary: `make build-tests`
-5) Run E2E Tests using the following command: `./bin/tests --ginkgo.v --ginkgo.progress --kubeconfig=/var/run/kubernetes/admin.kubeconfig`
 
+Once you have the driver installed, tests can be run with:
+``` sh
+$ make build-tests
+mkdir -p bin
+CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-X main.version=4fa924a251193c9eef937042112462433089d658 -extldflags "-static"' -o ./bin/tests ./cmd/tests
+$ ./bin/tests --ginkgo.v --ginkgo.progress
```
mathu97: do you know where the kubeconfig gets passed?

pohly: `KUBECONFIG` gets set by `start_cluster` based on the output of `kind get kubeconfig-path --name=csi-prow`.

It would be useful to make that more visible in the `prow.sh` output for those who try to reproduce what it does.

pohly: I've added two more commits which show KUBECONFIG (needs to be two separate commits because they'll go into different repos).

mathu97:
@pohly these are the pod logs i'm seeing for the `csi-nodeplugin-nfsplugin-` pod (`nfs` container)

```
E0614 19:26:01.292828       1 utils.go:50] GRPC error: rpc error: code = Internal desc = mount failed: exit status 32
Mounting command: mount
Mounting arguments: -t nfs 10.32.0.6:/ /var/lib/kubelet/pods/fe661032-8ed9-11e9-baf8-024275a11a61/volumes/kubernetes.io~csi/csi-nfsplugin-vjvvj/mount
Output: mount.nfs: rpc.statd is not running but is required for remote locking.
mount.nfs: Either use '-o nolock' to keep locks local, or start statd.
mount.nfs: an incorrect mount option was specified

E0614 19:26:43.582838       1 mount_linux.go:143] Mount failed: exit status 32
Mounting command: mount
Mounting arguments: -t nfs 10.32.0.6:/ /var/lib/kubelet/pods/fe661032-8ed9-11e9-baf8-024275a11a61/volumes/kubernetes.io~csi/csi-nfsplugin-vjvvj/mount
Output: mount.nfs: rpc.statd is not running but is required for remote locking.
mount.nfs: Either use '-o nolock' to keep locks local, or start statd.
mount.nfs: an incorrect mount option was specified
```

I tried the tests again against a` hack/local-up-cluster` cluster,  and they are working. Maybe there's a config that needs to be added to the KIND cluster?

msau42:
You may need to bundle nfs-utils into the driver image and start up the rpcbind and statd services

pohly:
Looks like the NFS CSI driver makes assumptions about the host it runs on which are a bit too broad: the host simply might not have the RPC services running.

This is tricky. If the CSI driver could ensure that it is the only user on the system of them, it could provide them. But if it isn't and something else starts up that relies on them, then restarting or uninstalling the driver will break those other users.

It probably would be possible to log into the KinD cluster and enable RPC, but then the NFS CSI driver needs to document that this is a requirement.

Issue #3, opened 2019-02-15
Add information to README for csi-driver-nfs


saad-ali:
Add the follow information:

* What version of CSI does this driver support
* What is the status of this driver (alpha).
* Which of the following (optional) features does this driver support:
   * dynamic provisioning
   * online resize or offline resize
   * block
   * snapshot
* Which AccessModes does this driver support?

/assign @prateekpandey14

msau42:
@mathu97 could you help update the readme?

mathu97:
@msau42 yes, will get on it.

PR #27, opened 2020-03-15
Add basic info to README about the plugin compatibility and features

wozniakjan:
**What type of PR is this?**
/kind cleanup

Informs about basic compatibility and feature set of the plugin.

Fixes: https://github.com/kubernetes-csi/csi-driver-nfs/issues/3

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,79 +1,71 @@
 # CSI NFS driver
 
-## Kubernetes
-### Requirements
+## Overview
 
-The folllowing feature gates and runtime config have to be enabled to deploy the driver
+This is a repository for [NFS](https://en.wikipedia.org/wiki/Network_File_System) [CSI](https://kubernetes-csi.github.io/docs/) Driver.
+Currently it implements bare minimum of the [CSI spec](https://github.com/container-storage-interface/spec) and is in the alpha state 
+of the development.
 
+#### CSI Feature matrix
+
+| **nfs.csi.k8s.io** | K8s version compatibility | CSI versions compatibility | Dynamic Provisioning | Resize | Snapshots | Raw Block | AccessModes              | Status                                                                       |
+|--------------------|---------------------------|----------------------------|----------------------|--------|-----------|-----------|--------------------------|------------------------------------------------------------------------------|
+|master              | 1.14 +                    | v1.0 +                     |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha                                                                        |
+|v2.0.0              | 1.14 +                    | v1.0 +                     |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha                                                                        |
+|v1.0.0              | 1.9 - 1.15                | v1.0                       |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | [deprecated](https://github.com/kubernetes-csi/drivers/tree/master/pkg/nfs)  |
+
+## Requirements
+
+The CSI NFS driver requires Kubernetes cluster of version 1.14 or newer and 
+preexisting NFS server, whether it is deployed on cluster or provisioned 
+independently. The plugin itself provides only a communication layer between 
+resources in the cluser and the NFS server.
+
+## Example
+
+There are multiple ways to create a kubernetes cluster, the NFS CSI plugin
+should work invariantly of your cluster setup. Very simple way of getting
+a local environment for testing can be achieved using for example
+[kind](https://github.com/kubernetes-sigs/kind).
+
+There are also multiple different NFS servers you can use for testing of 
+the plugin, the major versions of the protocol v2, v3 and v4 should be supported
+by the current implementation.
+
+The example assumes you have your cluster created (e.g. `kind create cluster`)
+and working NFS server (e.g. https://github.com/rootfs/nfs-ganesha-docker)
+
+#### Deploy
+
+Deploy the NFS plugin along with the `CSIDriver` info.
 ```
-FEATURE_GATES=CSIPersistentVolume=true,MountPropagation=true
-RUNTIME_CONFIG="storage.k8s.io/v1alpha1=true"
-```
-
-Mountprogpation requries support for privileged containers. So, make sure privileged containers are enabled in the cluster.
-
-### Example local-up-cluster.sh
-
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,6 +1,17 @@
...
+
+This is a repository for [NFS](https://en.wikipedia.org/wiki/Network_File_System) [CSI](https://kubernetes-csi.github.io/docs/) Driver.
+Currently it implements bare minimum of the CSI and the state is alpha.
+
+#### CSI Feature matrix
+
+| **nfs.csi.k8s.io** | CSI versions compatibility| Dynamic Provisioning | Resize | Snapshots | Raw Block | AccessModes              | Status |
+|--------------------|---------------------------|----------------------|--------|-----------|-----------|--------------------------|--------|
+|master              | v1.0.0, v1.1.0            |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha  |
+|v1.0.0              | v1.0.0, v1.1.0            |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha  |
```
msau42: We may have to go with a v1.1 or maybe even a 2.0? The old driver in the deprecated repo already has a 1.0.2: https://github.com/kubernetes-csi/drivers/releases

README.md
```diff
@@ -1,6 +1,17 @@
...
+
 ### Requirements
 
 The folllowing feature gates and runtime config have to be enabled to deploy the driver
```
msau42: I'm unsure if all the feature gates and alpha runtime config below are necessary anymore.

README.md
```diff
@@ -1,6 +1,17 @@
...
 
-## Kubernetes
+## Overview
+
+This is a repository for [NFS](https://en.wikipedia.org/wiki/Network_File_System) [CSI](https://kubernetes-csi.github.io/docs/) Driver.
+Currently it implements bare minimum of the CSI and the state is alpha.
+
+#### CSI Feature matrix
+
+| **nfs.csi.k8s.io** | CSI versions compatibility| Dynamic Provisioning | Resize | Snapshots | Raw Block | AccessModes              | Status |
```
msau42: Can you also add the minimum K8s version to this table? It is 1.14 due to the dependency on CSIDriver beta object.

README.md
```diff
@@ -1,6 +1,17 @@
 # CSI NFS driver
 
-## Kubernetes
+## Overview
+
+This is a repository for [NFS](https://en.wikipedia.org/wiki/Network_File_System) [CSI](https://kubernetes-csi.github.io/docs/) Driver.
```
msau42: Can you clarify somewhere near the top that the driver assumes the nfs server has already been deployed/available separately?

README.md
msau42: Also I think we should remove the csc tool instructions. I don't think it's been maintained for a while now.

wozniakjan:
very much agree with all your comments, will have them addressed by the end of weekend. Thanks for taking a thorough look at the docs @msau42  :)

wozniakjan:
@msau42, updated the docs per your suggestions, also briefly expanded the `example` section because without the CSC tool, it would be pretty much just the title of the section without any text :)

https://github.com/kubernetes-csi/csi-driver-nfs/blob/e2aa33b39982a5dee382fdf3ea75eb3ee13a09a0/README.md

README.md
```diff
@@ -1,79 +1,69 @@
...
-FEATURE_GATES=CSIPersistentVolume=true,MountPropagation=true
-RUNTIME_CONFIG="storage.k8s.io/v1alpha1=true"
-```
+#### CSI Feature matrix
 
-Mountprogpation requries support for privileged containers. So, make sure privileged containers are enabled in the cluster.
+| **nfs.csi.k8s.io** | Min K8s version | CSI versions compatibility | Dynamic Provisioning | Resize | Snapshots | Raw Block | AccessModes              | Status |
+|--------------------|-----------------|----------------------------|----------------------|--------|-----------|-----------|--------------------------|--------|
+|master              | 1.14            | v1.1                       |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha  |
+|v1.0.0              | 1.14            | v1.1                       |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha  |
```
msau42: Just a tiny nit, let's start the versioning at 2.0 to distinguish it from the old driver.

wozniakjan: sure, just so I understand what you mean, CSI version to bump to v2.0?

wozniakjan: maybe I can clarify the table a bit too to avoid any confusion, `master` == `v1.0.0`, there has not been a release of this yet. Theoretically we can start with `v2.0.0` if that is what you were referring to, but that means `v1.0.0` will probably never be released :)

msau42: The problem is we actually do have a 1.0 released, in the old repo: https://github.com/kubernetes-csi/drivers/releases, and we're going to reuse the the same image repo, so 1.0 is already taken.

wozniakjan: ah, now with the second repo it makes a lot of sense. My bad, adjusted the README and manifest + version in code to reflect the desired v2.0.0 release version. Thanks for clarifying that

README.md
```diff
@@ -1,87 +1,79 @@
...
+|--------------------|---------------------------|----------------------------|----------------------|--------|-----------|-----------|--------------------------|------------------------------------------------------------------------------|
+|master              | 1.14 +                    | v1.0 +                     |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha                                                                        |
+|v2.0.0              | 1.14 +                    | v1.0 +                     |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha                                                                        |
+|v1.0.0              | 1.9 - 1.15                | v1.0                       |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | [deprecated](https://github.com/kubernetes-csi/drivers/tree/master/pkg/nfs)  |
```
msau42: why is the max 1.15?

wozniakjan: https://github.com/kubernetes-csi/drivers/blob/337ec802362eb64fa9b6abd1a1b14ad997575d72/pkg/nfs/deploy/kubernetes/csi-nodeplugin-nfsplugin.yaml#L4

uses API group `apps` deprecated in 1.16

msau42:
/lgtm
/approve

Thanks!

PR #36, opened 2020-06-17
Allow to set custom permissions for the mounted folder

Fedosin:
**What type of PR is this?**

/kind feature

For RWX volume, kubelet does not perform recursive ownership/permission change. The heuristics that kubelet uses is being modified via - https://github.com/kubernetes/enhancements/issues/1682

Having said that, for RWX volumes which are made available via NFS protocol, using fsGroup is not recommended because if there are 2 pods that are trying to use same volume but with different fsGroup then one pod may lock out the other pod.

To avoid this, we must be able to set the folder permissions to 777.

```release-note
Added an optional cli parameter `--mount-permissions`, that allows to define custom permissions for the mounted folder. If the value is not specified, then default permissions will be kept.
```

### DIFF
--- a/cmd/nfsplugin/main.go
+++ b/cmd/nfsplugin/main.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/spf13/cobra"
 
@@ -29,6 +30,7 @@ import (
 var (
 	endpoint string
 	nodeID   string
+	perm     string
 )
 
 func init() {
@@ -55,6 +57,8 @@ func main() {
 	cmd.PersistentFlags().StringVar(&endpoint, "endpoint", "", "CSI endpoint")
 	cmd.MarkPersistentFlagRequired("endpoint")
 
+	cmd.PersistentFlags().StringVar(&perm, "mount-permissions", "", "mounted folder permissions")
+
 	cmd.ParseFlags(os.Args[1:])
 	if err := cmd.Execute(); err != nil {
 		fmt.Fprintf(os.Stderr, "%s", err.Error())
@@ -65,6 +69,18 @@ func main() {
 }
 
 func handle() {
-	d := nfs.NewNFSdriver(nodeID, endpoint)
+	// Converting string permission representation to *uint32
+	var parsedPerm *uint32
+	if perm != "" {
+		permu64, err := strconv.ParseUint(perm, 8, 32)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Incorrect mount-permissions value: %q", perm)
+			os.Exit(1)
+		}
+		permu32 := uint32(permu64)
+		parsedPerm = &permu32
+	}
+
+	d := nfs.NewNFSdriver(nodeID, endpoint, parsedPerm)
 	d.Run()
 }
--- a/pkg/nfs/nodeserver.go
+++ b/pkg/nfs/nodeserver.go
@@ -73,6 +73,12 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if ns.Driver.perm != nil {
+		if err := os.Chmod(targetPath, os.FileMode(*ns.Driver.perm)); err != nil {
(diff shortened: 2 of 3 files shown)


cmd/nfsplugin/main.go
```diff
@@ -55,6 +56,8 @@ func main() {
 	cmd.PersistentFlags().StringVar(&endpoint, "endpoint", "", "CSI endpoint")
 	cmd.MarkPersistentFlagRequired("endpoint")
 
+	cmd.PersistentFlags().Uint32Var(&perm, "mount-permissions", 0750, "mounted folder permissions")
```
jsafrane, line 59: Does `PersistentFlags` parse octal numbers correctly? If not, use string and parse it afterwards.
In addition, you should ensure that the value covers only chmod-able bits (owner/group/other + suid/sgid that may be useful).

pkg/nfs/nodeserver.go
```diff
@@ -40,7 +40,7 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	notMnt, err := ns.mounter.IsLikelyNotMountPoint(targetPath)
 	if err != nil {
 		if os.IsNotExist(err) {
-			if err := os.MkdirAll(targetPath, 0750); err != nil {
+			if err := os.MkdirAll(targetPath, os.FileMode(ns.Driver.perm)); err != nil {
```
jsafrane, line 43: I don't think this chmod is necessary.

pkg/nfs/nodeserver.go
```diff
@@ -73,6 +73,10 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if err := os.Chmod(targetPath, os.FileMode(ns.Driver.perm)); err != nil {
```
gnufied, line 76: it should be possible to opt-out from the chmod thing entirely and hence `ns.Driver.Perm` should default to `nil` may be?

pkg/nfs/nodeserver.go
Fedosin: done

cmd/nfsplugin/main.go
Fedosin: yeah, now this is a string that we parse later

pkg/nfs/nodeserver.go
Fedosin: agree. now if the value is not set, we omit the chmoding.

pkg/nfs/nodeserver.go
```diff
@@ -73,6 +74,17 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if ns.Driver.perm != "" {
+		perm, err := strconv.ParseUint(ns.Driver.perm, 8, 32)
```
jsafrane, line 78: Can you please convert it to uint just after parsing cmdline, so the driver exits with a clear error message? Reporting cmdline error in CSI error code looks weird.

cmd/nfsplugin/main.go
```diff
@@ -55,6 +56,8 @@ func main() {
 	cmd.PersistentFlags().StringVar(&endpoint, "endpoint", "", "CSI endpoint")
 	cmd.MarkPersistentFlagRequired("endpoint")
 
+	cmd.PersistentFlags().StringVar(&perm, "mount-permissions", "", "mounted folder permissions")
```
jsafrane, line 59: There should be either a default value or there should be no chmod if the value is empty.

Fedosin: Yes, if the value is empty, we don't perform chmod: https://github.com/kubernetes-csi/csi-driver-nfs/pull/36/files#diff-e0808273996894677ae8cffcc061ed60R77

pkg/nfs/nodeserver.go
Fedosin: done, it's in cmd now.

msau42:
This is fine for a short term mitigation, but I would like to see more investment on the Kubernetes side to solve this problem more holistically. 777 is not great from a security hardening perspective.

/approve

PR #40, opened 2020-07-29
Remove hostnetwork from plugin

msau42:
**What type of PR is this?**
/kind bug

Remove hostnetwork from the driver Daemonset so that the csi driver can resolve K8s dns names.

Also clean up the example to use claimRef instead of a selector which is not as exact in matching.

Fixes #

```release-note
Fixes issue where the driver could not mount a nfs server via Kubernetes dns service name.
```

### DIFF
--- a/deploy/kubernetes/csi-nodeplugin-nfsplugin.yaml
+++ b/deploy/kubernetes/csi-nodeplugin-nfsplugin.yaml
@@ -14,7 +14,6 @@ spec:
         app: csi-nodeplugin-nfsplugin
     spec:
       serviceAccount: csi-nodeplugin
-      hostNetwork: true
       containers:
         - name: node-driver-registrar
           image: quay.io/k8scsi/csi-node-driver-registrar:v1.0.2
--- a/examples/kubernetes/nginx.yaml
+++ b/examples/kubernetes/nginx.yaml
@@ -5,6 +5,9 @@ metadata:
   labels:
     name: data-nfsplugin
 spec:
+  claimRef:
+    name: data-nfsplugin
+    namespace: default
   accessModes:
   - ReadWriteMany
   capacity:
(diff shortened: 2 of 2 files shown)


ensonic:
Unfortunately this causes this problem for me:
```
    MountVolume.SetUp failed for volume "data" : kubernetes.io/csi: mounter.SetupAt failed: rpc error: code = Internal desc = mount failed: exit status 32
    Mounting command: mount
    Mounting arguments: -t nfs nfs-server.app-data-server.svc.cluster.local:/exports /var/lib/kubelet/pods/937d2e55-7f39-46a9-ab57-b31d197a2868/volumes/kubernetes.io~csi/data/mount
    Output: mount.nfs: rpc.statd is not running but is required for remote locking.
    mount.nfs: Either use '-o nolock' to keep locks local, or start statd.
    mount.nfs: mounting nfs-server.app-data-server.svc.cluster.local:/exports failed, reason given by server: No such file or directory
```
I guess the nfs-server was talking to rpc.statd though the hostnetwork.

msau42:
@ensonic can you try "/" as the export? That's what I used for my testing. (I tried "/exports" with and without hostnetwork and it didn't work)

ensonic:
Wow, that's it. We should probably update the example here too:
https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/examples/kubernetes/nginx.yaml#L22

I'd never have guessed that form the error. Thanks for the pointer.

wozniakjan:
doesn't the exported path depend on the NFS server configuration? I remember that when I tried the plugin with ganesha, I had to make sure [the configured export path](https://docs.gluster.org/en/latest/Administrator%20Guide/NFS-Ganesha%20GlusterFS%20Integration/#step-1) matched the `PV.Spec.CSI.VolumeAttributes.Share`. So perhaps we can document this better? I think changing the `/export` to `/` in example is fine, but may have similar problem when the NFS server is configured to export only certain paths and not entire root.

PR #53, opened 2020-10-07
feat: implement `CreateVolume`

### LINKED ISSUE #30: Support CreateVolume
A complete csi plugin needs to finish its controllerserver.go
https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/pkg/nfs/controllerserver.go.

We can use the code in https://github.com/kubernetes-incubator/external-storage/blob/master/nfs-client/cmd/nfs-client-provisioner/provisioner.go as comparison.

mayankshah1607:
Signed-off-by: Mayank Shah <mayankshah1614@gmail.com>

**What type of PR is this?**
/kind feature

Implements `CreateVolume`

Fixes #30

```release-note
Add implementation for `CreateVolume` and `DeleteVolume`
```


mayankshah1607:
@andyzhangx @msau42  I have made small modifications to @msau42 's work on implementing `CreateVolume`- https://github.com/msau42/csi-driver-nfs/commit/da0e3d682ea34fd8cbcdd042d0f88235dde61bb5
 Could you please take a look? If things look fine, I shall open a follow-up PR implementing `DeleteVolume` as well
Thanks

mayankshah1607:
As this PR introduces unit tests, I was wondering if we should enable them on Github Actions. We could move on to prow once we start adding more tests.

andyzhangx:
> As this PR introduces unit tests, I was wondering if we should enable them on Github Actions. We could move on to prow once we start adding more tests.

sure, go ahead

pkg/nfs/controllerserver.go
```diff
@@ -10,10 +15,95 @@ import (
...
+const (
+	idServer = iota
+	idBaseDir
+	idSubDir
+	totalIDElements // Always last
+)
+
 func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+	glog.V(4).Infof("CreateVolume called with request %+v", req)
```
andyzhangx, line 70: remove the logging since it's not necessary

pkg/nfs/controllerserver.go
```diff
@@ -10,10 +15,95 @@ import (
...
+		if err = cs.internalUnmount(ctx, nfsVol); err != nil {
+			glog.Warningf("failed to unmount nfs server: %v", err.Error())
+		}
+	}()
+
+	// Create subdirectory under base-dir
+	// TODO: revisit permissions
+	internalVolumePath := cs.getInternalVolumePath(nfsVol)
+	glog.V(4).Infof("Creating subdirectory at %v", internalVolumePath)
+	if err = os.Mkdir(internalVolumePath, 0755); err != nil {
```
andyzhangx, line 101: make default as 0777

andyzhangx (commented):
pls add `ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME` here:
https://github.com/kubernetes-csi/csi-driver-nfs/blob/7fdf12763de819564a19bffdc06c043fac7250a7/pkg/nfs/nfs.go#L72

also have you tried manually to make sure it works?

pkg/nfs/controllerserver.go
```diff
@@ -73,3 +161,177 @@ func (cs *ControllerServer) ListSnapshots(ctx context.Context, req *csi.ListSnap
...
+		subDir:  name,
+	}
+	vol.id = cs.getVolumeIdFromNfsVol(vol)
+
+	return vol, nil
+}
+
+// Get working directory for CreateVolume and DeleteVolume
+func (cs *ControllerServer) getInternalMountPath(vol *nfsVolume) string {
+	return filepath.Join(cs.workingMountDir, vol.subDir)
```
andyzhangx, line 284: what if `workingMountDir` is empty?

andyzhangx: it's better define a default path, e.g. `/mnt`

pkg/nfs/controllerserver.go
```diff
@@ -73,3 +161,177 @@ func (cs *ControllerServer) ListSnapshots(ctx context.Context, req *csi.ListSnap
...
+//     CreateVolume calls in parallel and they may use the same underlying share.
+//     Instead of refcounting how many CreateVolume calls are using the same
+//     share, it's simpler to just do a mount per request.
+func (cs *ControllerServer) getInternalVolumePath(vol *nfsVolume) string {
+	return filepath.Join(cs.getInternalMountPath(vol), vol.subDir)
+}
+
+// Get user-visible share path for the volume
+func (cs *ControllerServer) getVolumeSharePath(vol *nfsVolume) string {
+	return filepath.Join("/", vol.baseDir, vol.subDir)
```
andyzhangx, line 300: why this `filepath.Join` takes 3 parameters?

pkg/nfs/controllerserver.go
```diff
@@ -73,3 +161,177 @@ func (cs *ControllerServer) ListSnapshots(ctx context.Context, req *csi.ListSnap
...
+		AccessType: &csi.VolumeCapability_Mount{
+			Mount: &csi.VolumeCapability_MountVolume{},
+		},
+	}
+
+	glog.V(4).Infof("internally mounting %v:%v at %v", vol.server, sharePath, targetPath)
+	_, err := cs.Driver.ns.NodePublishVolume(ctx, &csi.NodePublishVolumeRequest{
+		TargetPath: targetPath,
+		VolumeContext: map[string]string{
+			"server": vol.server,
```
andyzhangx, line 223: use `paramServer`

pkg/nfs/controllerserver.go
```diff
@@ -73,3 +161,177 @@ func (cs *ControllerServer) ListSnapshots(ctx context.Context, req *csi.ListSnap
...
+			"share":  sharePath,
```
andyzhangx, line 224: use `paramShare`

mayankshah1607:
> pls add ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME here:

@andyzhangx On doing so, the sanity tests are failing with:
```
I1008 09:23:27.523659    4253 utils.go:48] GRPC call: /csi.v1.Controller/CreateVolume
I1008 09:23:27.524697    4253 utils.go:49] GRPC request: {"capacity_range":{"required_bytes":10737418240},"name":"sanity-controller-create-single-with-capacity-0E4531D2-F7FFA39B","volume_capabilities":[{"AccessType":{"Mount":{}},"access_mode":{"mode":1}}]}
E1008 09:23:27.543171    4253 utils.go:52] GRPC error: rpc error: code = InvalidArgument desc = server is a required parameter
```

pkg/nfs/controllerserver.go
msau42: This is the path of the share that was created and put into the returned VolumeContext. ie, if you have Storageclass param `base-dir = myvol`, `server = 10.10.10.10`, then after CreateVolume finishes, the VolumeContext returned should be: `server = 10.10.10.10`, `share = /myvol/foo-subdir`

msau42:
> As this PR introduces unit tests, I was wondering if we should enable them on Github Actions. We could move on to prow once we start adding more tests.

Unit tests should already be run as part of pull-kubernetes-csi-csi-driver-nfs. You can double check the build log of the job to see what gets tested

andyzhangx:
could you fix the sanity test failures? Thanks.

pkg/nfs/controllerserver.go
```diff
@@ -73,3 +161,166 @@ func (cs *ControllerServer) ListSnapshots(ctx context.Context, req *csi.ListSnap
...
+			return nil, fmt.Errorf("invalid parameter %q", k)
+		}
+	}
+
+	// Validate required parameters
+	if server == "" {
+		return nil, fmt.Errorf("%v is a required parameter", paramServer)
+	}
+	if baseDir == "" {
+		return nil, fmt.Errorf("%v is a required parameter", paramBaseDir)
```
mayankshah1607, line 269: @andyzhangx The sanity tests errors are coming here. If we want to keep this validation, we will have to update the existing sanity tests to use `server` and `baseDir` as request parameters. 
Is that ok?

pkg/nfs/controllerserver.go
```diff
@@ -10,10 +15,93 @@ import (
...
+	paramServer = "server"
+	// Base directory of the NFS server to create volumes under.
+	// The base directory must be a direct child of the root directory.
+	// The root directory is ommitted from the string, for example:
+	//     "base" instead of "/base"
+	paramBaseDir = "base-dir"
+)
+
+// CSI Volume attributes
+const (
```
andyzhangx, line 49: use one `const (` struct

pkg/nfs/controllerserver.go
```diff
@@ -10,10 +15,93 @@ import (
...
+	idServer = iota
+	idBaseDir
+	idSubDir
+	totalIDElements // Always last
+)
+
 func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+
```
andyzhangx, line 70: remove this empty line

pkg/nfs/controllerserver.go
andyzhangx: need to define a testing nfs-server in sanity test, otherwise even we ignore this error, `CreateVolume` would fail finally.

mayankshah1607: And what would the values of these parameters be? Would adding these to the [defaults](func NewTestConfig() TestConfig {) break any existing tests?

mayankshah1607: > need to set up a testing nfs-server in sanity test, otherwise even we ignore this error, CreateVolume would fail finally.

Can we run the [existing nfs server example](https://github.com/kubernetes-csi/csi-driver-nfs/tree/master/examples/kubernetes/nfs-provisioner) before initiating sanity tests? How would we feed in the `server` and `baseDir` parameters?

andyzhangx: I think yes, and I have modified the nfs server example: https://github.com/kubernetes-csi/csi-driver-nfs/pull/61, now could use `nfs-server.default.svc.cluster.local:/` in sanity test.

Need to add a flag in driver, if it's sanity test, use `nfs-server.default.svc.cluster.local:/`

andyzhangx: ah, in sanity test, there is no k8s cluster, it has only one machine, so you need to set up a containerized nfs-server, and use `127.0.0.1:/` for nfs-server.

mayankshah1607: Seems like there's an issue with running docker containers within the sanity-test container:
```
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
```
Is docker in docker the right way to go about provisioning an nfs server for sanity tests?

mayankshah1607:
@andyzhangx The sanity test is failing to start docker:
```
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
```

It looks like we're going to have to modify the present prow job to opt-in for Docker in Docker by adding the `preset-dind-enabled: "true` label. Example - https://github.com/kubernetes/test-infra/pull/9983/files
If this approach for running an NFS server is desirable, shall I open a PR against k/test-infra?

andyzhangx:
> @andyzhangx @msau42 The sanity test is failing to start docker:
> 
> ```
> docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
> ```
> 
> It looks like we're going to have to modify the present prow job to opt-in for Docker in Docker by adding the `preset-dind-enabled: "true"` label. Example - https://github.com/kubernetes/test-infra/pull/9983/files
> If this approach for running an NFS server is desirable, shall I open a PR against k/test-infra?

@mayankshah1607 what about run sanity test in github action, that could be easier to configure?

test/sanity/run-test.sh
```diff
@@ -32,6 +32,12 @@ function install_csi_sanity_bin {
   popd
 }
 
+function provision_nfs_server {
+  echo 'Installing NFS server on localhost'
+  docker run -d --name nfs --privileged -p 2049:2049 -v $(pwd):/nfsshare -e SHARED_DIRECTORY=/nfsshare itsthenetwork/nfs-server-alpine:latest
```
andyzhangx, line 37: may need 111 UDP port?

andyzhangx:
great job, only two sanity test failures left:
```
Summarizing 2 Failures:

[Fail] Controller Service [Controller Server] CreateVolume [It] should not fail when requesting to create a volume with already existing name and same capacity. 
/home/runner/go/pkg/mod/github.com/kubernetes-csi/csi-test@v2.2.0+incompatible/pkg/sanity/controller.go:567

[Fail] Controller Service [Controller Server] CreateVolume [It] should fail when requesting to create a volume with already existing name and different capacity. 
/home/runner/go/pkg/mod/github.com/kubernetes-csi/csi-test@v2.2.0+incompatible/pkg/sanity/controller.go:646
```

pkg/nfs/controllerserver.go
```diff
@@ -76,12 +79,31 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
+		capacity, ok := metrics.Capacity.AsInt64()
+		if !ok {
+			return nil, status.Errorf(codes.Internal, "failed to get capacity")
+		}
+
+		if capacity != reqCapacity {
+			return nil, status.Errorf(codes.AlreadyExists, "volume at this path with different size already exists.")
+		}
+
```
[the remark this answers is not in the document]
andyzhangx: I think we could skip this case since `CreateVolume` could not set the capacity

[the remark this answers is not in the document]
andyzhangx: use example here to skip this case: https://github.com/kubernetes-csi/csi-driver-smb/blob/master/test/sanity/run-test.sh#L47

msau42:
A disadvantage of Github actions is that the k8s merge bot doesn't block on it. Have we tried fixing the prow job config to enable docker in docker and that didn't work? I guess once question is how well running multiple nfs servers in different containers will work.

pkg/nfs/controllerserver.go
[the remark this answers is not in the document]
msau42: Is this a problem with the CSI spec or with the test case? Can we open bugs to track fixing either the spec or test to work for our use case, which is not unusual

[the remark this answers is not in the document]
mayankshah1607: I suspect that the problem could be due to the way this implementation of `CreateVolume` is performing this particular check (with the `volume` package) w.r.t the test. I'm not sure if my understanding is correct, but if the test isn't really provisioning a volume, should the `NewMetricStatFs` method expect to work correctly?

pkg/nfs/controllerserver.go
```diff
@@ -10,14 +15,134 @@ import (
...
+		}
+		return &csi.CreateVolumeResponse{}, nil
+	}
+
+	if err = os.Mkdir(internalVolumePath, 0777); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to make subdirectory: %v", err.Error())
+	}
+	// Remove capacity setting when provisioner 1.4.0 is available with fix for
+	// https://github.com/kubernetes-csi/external-provisioner/pull/271
+	return &csi.CreateVolumeResponse{Volume: cs.nfsVolToCSI(nfsVol, reqCapacity)}, nil
```
mayankshah1607, line 95: @msau42 I was also a little confused about this part. Should the response not contain information about capacity?

msau42: this was a workaround until the issue in external-provisioner was fixed. Now that it is fixed, I think we can return 0 capacity.

pkg/nfs/controllerserver.go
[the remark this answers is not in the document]
andyzhangx: while such `req >= actual size` implementation would still make sanity test fail, so let's disable this test for this driver

mayankshah1607:
@andyzhangx @msau42 It looks like the sanity tests are successfully passing after enabling DinD in Prow. Should we remove the GH Actions workflow or keep both for now for ensuring stability?

Also, returning `CapacityBytes` as 0 in `CreateVolumeResponse` causes 2 tests to fail:

```
Controller Service [Controller Server] CreateVolume 
  should return appropriate values SingleNodeWriter WithCapacity 1Gi Type:Mount

Controller Service [Controller Server] CreateVolume 
  should fail when requesting to create a volume with already existing name and different capacity.
```
We could either skip these as well, or return the the same value as the request

andyzhangx:
> should return appropriate values SingleNodeWriter WithCapacity 1Gi Type:Mount

let's keep the GH action since it won't bring any more effort.
And about case `should return appropriate values SingleNodeWriter WithCapacity 1Gi Type:Mount`, is it possible to return same `CapacityBytes` as original request?

mayankshah1607:
> And about case should return appropriate values SingleNodeWriter WithCapacity 1Gi Type:Mount, is it possible to return same CapacityBytes as original request?

Yes, that's what I have done right now for the tests to pass. Just wanted to confirm if this would be a desirable behavior of the client, so we can proceed with getting this one merged if everything looks okay

.github/workflows/sanity.yaml
```diff
@@ -11,9 +11,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-      - name: Install nfs-common
```
andyzhangx, line 14: we don't need this change?

pkg/nfs/controllerserver.go
```diff
@@ -10,14 +15,134 @@ import (
...
+			glog.Warningf("failed to unmount nfs server: %v", err.Error())
+		}
+	}()
+
+	// Create subdirectory under base-dir
+	// TODO: revisit permissions
+	internalVolumePath := cs.getInternalVolumePath(nfsVol)
+
+	if volumePathExists(internalVolumePath) {
+		if err := validCapacity(reqCapacity, internalVolumePath); err != nil {
```
andyzhangx, line 84: is `validCapacity` still necessary since some sanity tests are not running any more, also the capacity is actually not controlled by this driver.

pkg/nfs/utils.go
```diff
@@ -16,9 +17,10 @@ func NewDefaultIdentityServer(d *nfsDriver) *IdentityServer {
 	}
 }
 
-func NewControllerServer(d *nfsDriver) *ControllerServer {
+func NewControllerServer(d *nfsDriver, workingDir string) *ControllerServer {
 	return &ControllerServer{
-		Driver: d,
+		Driver:          d,
+		workingMountDir: workingDir,
```
andyzhangx, line 23: in current implementation, `workingMountDir` is always `/tmp`, so maybe we don't need this field, just use a const var `workingMountDir` is enough

.github/workflows/sanity.yaml
mayankshah1607: This change was required so that the sanity test prow job could reuse this functionality

pkg/nfs/controllerserver.go
mayankshah1607: yes, I will remove it then

mayankshah1607:
@andyzhangx I have made these changes. PTAL, thanks!

pkg/nfs/utils.go
msau42: I believe it was parameterized this way to facilitate unit tests

pkg/nfs/controllerserver.go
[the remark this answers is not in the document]
msau42: I'm fine with disabling the test to unblock work, but we need to follow up with correcting the spec/test/driver. The test is supposed to make sure drivers are CSI compliant so if it's not working, we need to figure out why.

examples/kubernetes/storageclass-nfs.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: nfs-csi
+provisioner: nfs.csi.k8s.io
+parameters:
+  server: nfs-server.default.svc.cluster.local
+  share: /
+reclaimPolicy: Retain  # only retain is supported
```
msau42, line 10: Delete is supported now right?

mayankshah1607: Yes

andyzhangx (commented):
overall LGTM, thanks for the contribution!

andyzhangx (approved):
/lgtm

examples/kubernetes/storageclass-nfs.yaml
andyzhangx: pls fix this later.

andyzhangx (approved):
/approve

PR #61, opened 2020-10-12
doc: update NFS provisioner example

andyzhangx:
**What type of PR is this?**
/kind documentation

doc: update NFS provisioner example

Fixes #

```release-note
none
```


deploy/kubernetes/csi-nfs-node.yaml
```diff
@@ -14,6 +14,8 @@ spec:
       labels:
         app: csi-nfs-node
     spec:
+      hostNetwork: true
```
msau42, line 17: Why is host network needed? I had explicitly removed hostNetwork so that you could give an incluster hostname as the nfs server.

andyzhangx: `hostNetwork: true` is a must, I have tried after node driver daemonset  restarted, if there is no `hostNetwork: true` setting, original nfs connection would be broken. Same issue as https://github.com/kubernetes-csi/csi-driver-smb/issues/130

andyzhangx: @msau42 PTAL, thanks.

examples/kubernetes/nfs-provisioner/nfs-server.yaml
```diff
@@ -1,39 +1,59 @@
...
 kind: Service
 apiVersion: v1
 metadata:
-  name: nfs-service
+  name: nfs-server
+  labels:
+    app: nfs-server
 spec:
-  clusterIP: 10.0.171.239
+  type: ClusterIP  # use "LoadBalancer" to get a public ip
```
msau42, line 9: does this work in a kind cluster?

deploy/kubernetes/csi-nfs-node.yaml
msau42: thanks for the explanation, this makes sense. Can you add a comment here?

examples/kubernetes/nfs-provisioner/nfs-server.yaml
andyzhangx: `ClusterIP` type should always work, anyway it's better than a fixed IP config.

deploy/kubernetes/csi-nfs-node.yaml
andyzhangx: added a comments in yaml config, PTAL.

examples/kubernetes/nfs-provisioner/README.md
```diff
@@ -8,11 +8,7 @@
 kubectl create -f https://raw.githubusercontent.com/kubernetes-csi/csi-driver-nfs/master/examples/kubernetes/nfs-provisioner/nfs-server.yaml
```
msau42, line 8: One more nit, can we name this directory "nfs-server" instead so we don't confuse it with dynamic provisioning?

andyzhangx: fixed, PTAL

msau42:
/lgtm 

thanks!

PR #92, opened 2020-11-17
test: add more test cases for e2e test

songjiaxun:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind failing-test

Add more test cases in e2e test, including install scripts tests and log collection.
Add more example deployments.
**NOTE:** the changes require documentation changes. Will do it in another PR.

Fixes #

```release-note
None
```


examples/deployment-readwriteonce.yaml
```diff
@@ -0,0 +1,43 @@
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: pvc-deployment-nfs-rwo
+spec:
+  accessModes:
+    - ReadWriteOnce
```
andyzhangx, line 8: I don't think we need two similar yaml files, one for `ReadWriteOnce` and one for `ReadWriteMany`, I would suggest only keep `ReadWriteMany`, and add comment in that yaml file, e.g. `Available values: ReadWriteMany, ReadWriteOnce, ReadOnlyMany`, actually all `ReadWriteMany` accessModes would also those 3 modes

songjiaxun: OK, agree. Will make the change.

songjiaxun:
The e2e test passed, but error messages are found in the log. Will trouble it as well in this PR.
image

songjiaxun:
> The e2e test passed, but error messages are found in the log. Will troubleshoot it as well in this PR.
> image

Actually, it cannot be fixed because `nfsDriver` must be created before `flag.Parse()` is called.
https://github.com/kubernetes-csi/csi-driver-nfs/blob/d325ee36adc233c0f3260906b545bda3802db4c9/test/e2e/e2e_suite_test.go#L30
Because `type nfsDriver` is defined as a private type, and can only be created by a "constructor".
https://github.com/kubernetes-csi/csi-driver-nfs/blob/d325ee36adc233c0f3260906b545bda3802db4c9/pkg/nfs/nfs.go#L25

andyzhangx (approved):
/lgtm
/approve

PR #97, opened 2020-11-18
fix: make verify checks

### LINKED ISSUE #93: add livenessprobe for driver node deployment

could refer to https://github.com/kubernetes-csi/csi-driver-smb/blob/de292c34a09dd3e5c342b232e5f89c90137643c7/deploy/csi-smb-node.yaml#L24-L40

**Additional context**

songjiaxun:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind failing-test

Fix all the issues `make verify` may have, including removing unnecessary unused dependencies, adding headers, etc.
Add liveness probe for node server. https://github.com/kubernetes-csi/csi-driver-nfs/issues/93
After this change, `make verify`, `make sanity-test`, `make integration-test`, `make e2e-test` should all work. @andyzhangx Please help add "verify" and "integration test" in Github CI.

Fixes #93

```release-note
None
```


andyzhangx (approved):
/lgtm
/approve

andyzhangx:
@mayankshah1607 could you help add unit-test, integration-test to the test-infra, add to:
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-csi/csi-driver-nfs/csi-driver-nfs-unmanaged.yaml

And you could refer to smb csi driver config, thanks.

andyzhangx:
@songjiaxun could you fix the test error? thanks.

songjiaxun:
> @songjiaxun could you fix the test error? thanks.

I reverted the changes to `release-tools`. Verification tests under `hack` detected some issues in `release-tools`, however, the tests in `release-tools` itself do not allow changes from outside. This PR will be all right, but when we enable tests in `hack`, it will be a problem. It looks like there is an overlap between `release-tools` and `hack`. As @msau42 mentioned, I plan to combine them together. @mayankshah1607 when you enable the unit test in test-infra, please just use `make unit-test` instead of `make verify` as `release-tools` will do the verify work. Thank you!

andyzhangx (approved):
/lgtm
/approve

andyzhangx:
> make verify

@songjiaxun `make verify` should also work right? it's ok if we run redundant CI tests

songjiaxun:
> > make verify
> 
> @songjiaxun `make verify` should also work right? it's ok if we run redundant CI tests

I guess `make verify` would fail because some scripts in `release-tools` have misspelled words and some of the scripts do not have proper headers, which will cause the `misspelling check` and `boilerplate check` to fail. These cannot be fixed in this repo because `release-tools` has one step to check if itself has been modified from outside of its repo. Basically, the checks under `hack` and `release-tools` cannot exist at the same time I believe.

pohly:
> the tests in release-tools itself do not allow changes from outside.

The tests themselves cannot be *changed*, but that shouldn't be needed because all projects are expected to pass them. This bar isn't set too high because we want to be sure that all projects can achieve that without too much work.

You can *extend* them by adding additional make targets in your top-level Makefile.

> I guess make verify would fail because some scripts in release-tools have misspelled words and some of the scripts do not have proper headers, which will cause the misspelling check and boilerplate check to fail. 

If a project has stricter requirements than others, then it must exclude the "release-tools" directory from its own checks because there's no guarantee that "release-tools" will pass them.

PR #110, opened 2020-12-01
Update vendor to k8s 1.19 and csi-lib-utils to 0.9.0

msau42:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

 /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Updates csi-lib-utils to 0.9.0 to switch to klog v2

Fixes #

```release-note
Switching to klog v2
```


msau42:
/assign @jsafrane @pohly 

I don't know why, but klog v1 is still in the go.sum even though all the code and vendor/ are using klog v2 now. "go mod tidy" didn't help

pohly:
> I don't know why, but klog v1 is still in the go.sum even though all the code and vendor/ are using klog v2 now. 

I believe go.sum keeps some historic information. For example, `k8s.io/klog v0.3.0` also has an entry. Not exactly sure though when that gets purged. Removing `go.sum` and re-creating it drops some entries, but not `k8s.io/klog v0.3.0`. I think we can ignore that.

What matters is that `vendor/k8s.io/klog/` now only contains `v2`, so:
/lgtm

Issue #128, opened 2021-01-03
enable spelling and boilerplate check 


andyzhangx:
**Is your feature request related to a problem?/Why is this needed**

**Describe the solution you'd like in detail**

https://github.com/kubernetes-csi/csi-driver-nfs/pull/127 disabled following two checks:
 - verify-spelling.sh
 - verify-boilerplate.sh

need to upstream the above two scripts to https://github.com/kubernetes-csi/csi-release-tools.git and then re-enable these two scripts in this project

**Describe alternatives you've considered**

**Additional context**

```
error: docs/csi-dev.md:38:12: "environmnet" is a misspelling of "environments"
error: docs/csi-dev.md:89:6: "environmnet" is a misspelling of "environments"
error: release-tools/SIDECAR_RELEASE_PROCESS.md:14:17: "maintainence" is a misspelling of "maintenance"
error: release-tools/prow.sh:76:31: "seperately" is a misspelling of "separately"
error: release-tools/prow.sh:296:37: "seperated" is a misspelling of "separated"
error: release-tools/prow.sh:1001:32: "succesful" is a misspelling of "successful"
Found spelling errors!
Cleaning up...
Makefile:46: recipe for target 'verify' failed
make: *** [verify] Error 1
Error: Process completed with exit code 2.
```

andyzhangx:
@sachinkumarsingh092 could you help? thanks.

sachinkumarsingh092:
So we just have to push the `hack/boilerplate`, `hack/verify-spellcheck.sh` and `hack/boilerplate.sh` to https://github.com/kubernetes-csi/csi-release-tools.git, modify the `build.make` there to include these tests and after this is merged we can `git subtree pull` it for out repo. It this correct or am I missing something?

/assign

andyzhangx:
> So we just have to push the `hack/boilerplate`, `hack/verify-spellcheck.sh` and `hack/boilerplate.sh` to https://github.com/kubernetes-csi/csi-release-tools.git, modify the `build.make` there to include these tests and after this is merged we can `git subtree pull` it for this repo. Is this correct or am I missing something?
> 
> /assign

correct, thanks.

PR #133, opened 2021-01-12
Update release tools

sachinkumarsingh092:
**What type of PR is this?**
/kind cleanup

Updates the release tools for this repository after which we can enable spelling and boilerplate checks.

Part of #128 

```release-note
none
```

### DIFF
--- a/release-tools/.prow.sh
+++ b/release-tools/.prow.sh
@@ -1,7 +1,23 @@
 #! /bin/bash -e
+
+# Copyright 2021 The Kubernetes Authors.
 #
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 # This is for testing csi-release-tools itself in Prow. All other
 # repos use prow.sh for that, but as csi-release-tools isn't a normal
 # repo with some Go code in it, it has a custom Prow test script.
 
 ./verify-shellcheck.sh "$(pwd)"
+./verify-spelling.sh "$(pwd)"
+./verify-boilerplate.sh "$(pwd)"
--- a/release-tools/build.make
+++ b/release-tools/build.make
@@ -275,3 +275,16 @@ test-shellcheck:
 .PHONY: check-go-version-%
 check-go-version-%:
 	./release-tools/verify-go-version.sh "$*"
+
+# Test for spelling errors.
+.PHONY: test-spelling
+test-spelling:
+	@ echo; echo "### $@:"
+	@ ./release-tools/verify-spelling.sh "$(pwd)"
+
+# Test the boilerplates of the files.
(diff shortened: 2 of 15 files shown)


sachinkumarsingh092:
The failed test exits with:
```
./release-tools/verify-spelling.sh: line 52: cd: ./release-tools/..: No such file or directory
```
but the path does exist! Does it need an absolute path then?

release-tools/build.make
```diff
@@ -275,3 +275,18 @@ test-shellcheck:
 .PHONY: check-go-version-%
 check-go-version-%:
 	./release-tools/verify-go-version.sh "$*"
+
+# Test for spelling errors.
+.PHONY: test-spelling
+test: test-spelling
+test-spelling:
+	@ echo; echo "### $@:"
+	@ ./release-tools/verify-spelling.sh
```
pohly, line 284: After the change discussed in https://github.com/kubernetes-csi/csi-release-tools/pull/124#pullrequestreview-565256413 this invocation needs the absolute path of the directory which is to be checked here.

I didn't know that we were enabling this for *all* repos. This part here wasn't in PR that I reviewed (https://github.com/kubernetes-csi/csi-release-tools/pull/124/), so I missed this.

Is that really what we want? It will cause a lot of work the next time someone tries to update csi-release-tools in a repo because in addition to updating csi-release-tools, also all new errors found by this check will have to be fixed.

I would prefer to keep this check optional at first, i.e. *not* make it a dependency of `test` in `build.make`. If you want to activate it in csi-driver-nfs, you can put that dependencies into your top-level Makefile.

@msau42 what do you think?

release-tools/build.make
```diff
@@ -275,3 +275,18 @@ test-shellcheck:
...
+
+# Test the boilerplates of the files.
+.PHONY: test-boilerplate
+test: test-boilerplate
+test-boilerplate:
+	@ echo; echo "### $@:"
+	@ ./release-tools/verify-boilerplate.sh
```
pohly, line 291: Same here.

release-tools/build.make
pohly: In addition to passing `"$(pwd)"` here (for the sake of clarity) it would be good to also make the default root an absolute path (in case someone invokes the script directly without parameters).

msau42: I think we should have it enabled in all repos, but making it optional so we can roll it out more gradually is fine.

sachinkumarsingh092: Done!

release-tools/.prow.sh
```diff
@@ -1,7 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 # This is for testing csi-release-tools itself in Prow. All other
 # repos use prow.sh for that, but as csi-release-tools isn't a normal
 # repo with some Go code in it, it has a custom Prow test script.
 
 ./verify-shellcheck.sh "$(pwd)"
+./verify-spelling.sh "$(pwd)"
```
msau42, line 22: Is the goal to also enable these checks in the nfs repo? If so, we also need to add these calls to this repo's .prow.sh

sachinkumarsingh092: > Is the goal to also enable these checks in the nfs repo?

Yes, but this PR is solely to update the release tools for nfs repo. I intend to make a separate PR for enabling these tests for the prow and in the makefile (by enabling them in `hack/verify-all.sh`) hence fixing #128. Is it okay?

msau42: sounds good thanks!

PR #161, opened 2021-02-17
doc: update supported k8s version

songjiaxun:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind documentation
Update k8s support version

Fixes #

```release-note
none
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ of the development.
 
 | **nfs.csi.k8s.io** | K8s version compatibility | CSI versions compatibility | Dynamic Provisioning | Resize | Snapshots | Raw Block | AccessModes              | Status                                                                       |
 |--------------------|---------------------------|----------------------------|----------------------|--------|-----------|-----------|--------------------------|------------------------------------------------------------------------------|
-|master              | 1.14 +                    | v1.0 +                     |  yes                 |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha                                                                        |
+|master              | 1.16 +                    | v1.0 +                     |  yes                 |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha                                                                        |
 |v2.0.0              | 1.14 +                    | v1.0 +                     |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha                                                                        |
 
 ### Requirements


charts/README.md
```diff
@@ -6,13 +6,13 @@
 ## install latest version
 ```console
 helm repo add csi-driver-nfs https://raw.githubusercontent.com/kubernetes-csi/csi-driver-nfs/master/charts
-helm install csi-driver-nfs csi-driver-nfs/csi-driver-nfs --namespace kube-system
+helm install csi-driver-nfs csi-driver-nfs/csi-driver-nfs --namespace kube-system --disable-openapi-validation
```
andyzhangx, line 9: may add a note here, it's not necessary to make it as default

deploy/install-driver.sh
```diff
@@ -35,7 +35,7 @@ fi
 
 echo "Installing NFS CSI driver, version: $ver ..."
 kubectl apply -f $repo/rbac-csi-nfs-controller.yaml
-kubectl apply -f $repo/csi-nfs-driverinfo.yaml
+kubectl apply -f $repo/csi-nfs-driverinfo.yaml --validate=false
```
andyzhangx, line 38: it's not necessary to make it as default, remove this

Makefile
```diff
@@ -114,9 +114,10 @@ e2e-bootstrap: install-helm
 	helm install csi-driver-nfs ./charts/latest/csi-driver-nfs --namespace kube-system --wait --timeout=15m -v=5 --debug \
 	--set image.nfs.repository=$(REGISTRY)/$(IMAGENAME) \
 	--set image.nfs.tag=$(IMAGE_VERSION) \
-	--set image.nfs.pullPolicy=Always
-	--set controller.logLevel=8
-	--set node.logLevel=8
+	--set image.nfs.pullPolicy=Always \
+	--set controller.logLevel=8 \
+	--set node.logLevel=8 \
+	--disable-openapi-validation
```
andyzhangx, line 120: it's not necessary to make it as default, only enable this if `disable-openapi-validation` flag is defined.

deploy/install-driver.sh
songjiaxun: Actually, in our e2e test, this script is being called. And in Azure Stack CI, we are testing against Kubernetes v1.15, so it's necessary for older Kubernete versions, especially on the home page we declare that we support 1.14+. I can add a flag here.

songjiaxun: In my opinion, it's benign to add this as default, or we can change the home page to say we only support k8s 1.16+. Otherwise, this script will not work for k8s older than 1.16.

andyzhangx: it's a burden to support 1.15 since 1.15 is already not supported by upstream, there could be new features later that could easily break this. If user still wants to use 1.15, I think provide helm installation is enough, while we cannot guarantee it's fully supported.

songjiaxun: Ok make sense, so let me change the support version to 1.16+ on home page? Or we leave 1.14+ there and add notes to disable the api validation for older versions?

andyzhangx: yes, thanks. We don't support old versions officially.

README.md
```diff
@@ -11,8 +11,8 @@ of the development.
 
 | **nfs.csi.k8s.io** | K8s version compatibility | CSI versions compatibility | Dynamic Provisioning | Resize | Snapshots | Raw Block | AccessModes              | Status                                                                       |
 |--------------------|---------------------------|----------------------------|----------------------|--------|-----------|-----------|--------------------------|------------------------------------------------------------------------------|
-|master              | 1.14 +                    | v1.0 +                     |  yes                 |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha                                                                        |
-|v2.0.0              | 1.14 +                    | v1.0 +                     |  no                  |  no    |  no       |  no       | Read/Write Multiple Pods | Alpha                                                                        |
```
andyzhangx, line 15: only need to update master branch version

andyzhangx (approved):
/ok-to-test
/lgtm
/approve

PR #186, opened 2021-04-11
Allow shares that are not at the root

### LINKED ISSUE #184: Problems when using share with multiple folders
**What happened**:
There are errors in the logs like the following when using a share in a storage class that contains multiple folders. It causes for example the following message:
controllerserver.go:120] failed to get nfs volume for volume id 10.128.134.206/data/kubernetes/k8sdev/pvc-8b8929dc-d172-441b-b4ce-2659e5906ba0 deletion: volume id "10.128.134.206/data/kubernetes/k8sdev/pvc-8b8929dc-d172-441b-b4ce-2659e5906ba0" unexpected format: got 5 token(s) instead of 3

Deletion of the pvc folders on the share is also not working, but I haven't found an error message for it. Maybe it is not implemented

Shares which contain multiple folders work without errors.

**How to reproduce it**:
Use a storageclass with the following parameter and create a pvs using that storage class

parameters:
  server: 10.128.134.206
  share: /data/kubernetes/k8sdev

I tried setting mount-permissions on the volumes but that didn't work which is the reason I looked in the logs. Mounting the volumes in pods work as expected. I expect for now that this causes the problems with the permissions not being applied, but I haven't looked through the code yet to be sure.

**Environment**:
- CSI Driver version: Build Date: "2021-02-21T11:44:03Z" Driver Version: v3.0.0 Git Commit: ac023bf4afed00a8be9609979f22144089fca721
- Kubernetes version (use `kubectl version`): 1.20
- OS (e.g. from /etc/os-release): Debian 10
- Kernel (e.g. `uname -a`): 4.19

juergenthomann:
**What type of PR is this?**
/kind bug

Currently the split in getNfsVolFromID expects that the share is just one folder like /data and not something like /data/kubernetes and fails in the later case with an unexpected amount of tokens.

This splits the id into server until the first / and subdir starting from the last / character. Everything between them is considered a part of baseDir

Fixes #184

```release-note
NONE
```

### DIFF
--- a/pkg/nfs/controllerserver.go
+++ b/pkg/nfs/controllerserver.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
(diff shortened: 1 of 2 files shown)


andyzhangx (commented):
/ok-to-test

pkg/nfs/controllerserver.go
```diff
@@ -359,15 +360,16 @@ func (cs *ControllerServer) getVolumeIDFromNfsVol(vol *nfsVolume) string {
 
 // Given a CSI volume id, return a nfsVolume
 func (cs *ControllerServer) getNfsVolFromID(id string) (*nfsVolume, error) {
-	tokens := strings.Split(id, "/")
-	if len(tokens) != totalIDElements {
-		return nil, fmt.Errorf("volume id %q unexpected format: got %v token(s) instead of %v", id, len(tokens), totalIDElements)
+	volRegex := regexp.MustCompile("^([^/]+)/(.*)/([^/]+)$")
```
andyzhangx, line 363: could you also write a ut for this `getNfsVolFromID` function? thanks.

juergenthomann: I added now some test cases for this function. As I never wrote much in go before so I hope the test is OK. If you think that some changes would be good, let me know and I will adjust the test.

andyzhangx (approved):
/lgtm
/approve
thanks, it's already good enough

PR #206, opened 2021-05-29
feat: support fsGroupPolicy feature

andyzhangx:
**What type of PR is this?**
/kind feature

feat: support fsGroupPolicy feature

Fixes #

```release-note
feat: support fsGroupPolicy feature
```


charts/README.md
```diff
@@ -3,6 +3,11 @@
 ## Prerequisites
  - [install Helm](https://helm.sh/docs/intro/quickstart/#install-helm)
 
+### Tips
+ - `--set controller.runOnMaster=true` could make csi-nfs-controller only run on master node
+ - `--set feature.enableFSGroupPolicy=true` could enable `fsGroupPolicy` on a k8s 1.20+ cluster
```
msau42, line 8: Note that this feature went beta in 1.22. Before that, it is alpha and is not enabled by default in Kubernetes.

Also please add warnings in the documentation that if they want to use this feature, they need to be careful and make sure multiple Pods sharing the same volume all have the same fsgroup. Otherwise the file permissions can change while the volume is still mounted by other pods.

andyzhangx: already added comments:

 - `--set feature.enableFSGroupPolicy=true` could enable `fsGroupPolicy` on a k8s 1.20+ cluster (this feature is in beta, check details [here](../deploy/example/fsgroup))

PR #226, opened 2021-09-19
feat: ReadWriteOncePod support

andyzhangx:
**What type of PR is this?**
/kind feature

 - chore: upgrade to csi spec v1.5.0
 - feat: ReadWriteOncePod support which is alpha feature in k8s 1.22

Fixes #

```release-note
feat: ReadWriteOncePod support
```


chrishenzie (changes requested):
In order to provide full support for this feature, you will also need to [advertise the `SINGLE_NODE_MULTI_WRITER` node and controller capabilities here](https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/pkg/nfs/nfs.go#L80-L90).

These capabilities are required because CSI clients like the external-provisioner, attacher, and kubelet use them to determine which mapping of Kubernetes access modes to CSI access modes to use. Please see [this part of the feature blog](https://kubernetes.io/blog/2021/09/13/read-write-once-pod-access-mode-alpha/#advertise-new-csi-controller-and-node-service-capabilities) for more details, and [here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/2485-read-write-once-pod-pv-access-mode/README.md#csi-specification-changes-volume-capabilities) for more context on the motivations.

As a stretch goal we will eventually want to [implement the `NodePublishVolume` behavior](https://kubernetes.io/blog/2021/09/13/read-write-once-pod-access-mode-alpha/#implement-nodepublishvolume-behavior) however this behavior is not currently asserted in the CSI sanity test suite. Once it's added we should update the test suite again and add support.

Let me know if you have any questions, thanks again for taking a look at this!

andyzhangx:
> In order to provide full support for this feature, you will also need to [advertise the `SINGLE_NODE_MULTI_WRITER` node and controller capabilities here](https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/pkg/nfs/nfs.go#L80-L90).
> 
> These capabilities are required because CSI clients like the external-provisioner, attacher, and kubelet use them to determine which mapping of Kubernetes access modes to CSI access modes to use. Please see [this part of the feature blog](https://kubernetes.io/blog/2021/09/13/read-write-once-pod-access-mode-alpha/#advertise-new-csi-controller-and-node-service-capabilities) for more details, and [here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/2485-read-write-once-pod-pv-access-mode/README.md#csi-specification-changes-volume-capabilities) for more context on the motivations.
> 
> As a stretch goal we will eventually want to [implement the `NodePublishVolume` behavior](https://kubernetes.io/blog/2021/09/13/read-write-once-pod-access-mode-alpha/#implement-nodepublishvolume-behavior) however this behavior is not currently asserted in the CSI sanity test suite. Once it's added we should update the test suite again and add support.
> 
> Let me know if you have any questions, thanks again for taking a look at this!

@chrishenzie thanks for the review, I have added `SINGLE_NODE_MULTI_WRITER cap for both node and controller. Agree that we should add sanity test for this feature in the future.

chrishenzie:
@andyzhangx Just checking, were you able to smoke test this driver in a cluster with the feature enabled and the right sidecars? I may be able to test myself but I'll need some time before I can get to it

andyzhangx:
disable GetCapabilities sanity test temporally, need to upgrade csi-test later:
```
Summarizing 2 Failures:
[Fail] Controller Service [Controller Server] ControllerGetCapabilities [It] should return appropriate capabilities 
/home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:151
[Fail] Node Service NodeGetCapabilities [It] should return appropriate capabilities 
/home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:345
Ran 28 of 78 Specs in 87.955 seconds
FAIL! -- 26 Passed | 2 Failed | 1 Pending | 49 Skipped
```

andyzhangx:
> @andyzhangx Just checking, were you able to smoke test this driver in a cluster with the feature enabled and the right sidecars? I may be able to test myself but I'll need some time before I can get to it

@chrishenzie not yet, while I could try this feature later, this PR just adds the ReadWriteOncePod cap first.

andyzhangx:
BTW, github action unit test failure is expected since https://coveralls.io/ is down for one day.

chrishenzie:
@andyzhangx The feature should work with these changes (implementing NodePublishVolume behavior is more for adhering to the CSI spec than it is a requirement for ReadWriteOncePod to work)

andyzhangx:
let's merge first

PR #316, opened 2022-04-19
fix: CVE-2022-1271 in image build

andyzhangx:
**What type of PR is this?**
/kind bug

fix: CVE-2022-1271 in image build
```
+----------+------------------+----------+-------------------+-------------------+--------------------------------------+
| LIBRARY  | VULNERABILITY ID | SEVERITY | INSTALLED VERSION |   FIXED VERSION   |                TITLE                 |
+----------+------------------+----------+-------------------+-------------------+--------------------------------------+
| gzip     | CVE-2022-1271    | HIGH     | 1.10-4            | 1.10-4+deb11u1    | gzip: arbitrary-file-write           |
|          |                  |          |                   |                   | vulnerability                        |
|          |                  |          |                   |                   | -->avd.aquasec.com/nvd/cve-2022-1271 |
+----------+                  +          +-------------------+-------------------+                                      +
| liblzma5 |                  |          | 5.2.5-2           | 5.2.5-2.1~deb11u1 |                                      |
|          |                  |          |                   |                   |                                      |
|          |                  |          |                   |                   |                                      |
+----------+------------------+----------+-------------------+-------------------+--------------------------------------+
```

Fixes #

**Requirements**:
- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Release note**:
```
fix: CVE-2022-1271 in image build
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM k8s.gcr.io/build-image/debian-base:bullseye-v1.1.0
+FROM k8s.gcr.io/build-image/debian-base:bullseye-v1.2.0
 
 ARG ARCH
 ARG binary=./bin/${ARCH}/nfsplugin
@@ -21,6 +21,6 @@ COPY ${binary} /nfsplugin
 RUN apt update && apt-mark unhold libcap2
 RUN clean-install ca-certificates mount nfs-common netbase
 # install updated packages to fix CVE issues
-RUN clean-install libgmp10 bsdutils libssl1.1 openssl libc6 libc-bin libsystemd0 libudev1 zlib1g
+RUN clean-install zlib1g gzip liblzma5
 
 ENTRYPOINT ["/nfsplugin"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,9 +18,7 @@ ARG ARCH
 ARG binary=./bin/${ARCH}/nfsplugin
 COPY ${binary} /nfsplugin
 
-RUN apt update && apt-mark unhold libcap2
+RUN apt update && apt upgrade -y && apt-mark unhold libcap2
 RUN clean-install ca-certificates mount nfs-common netbase
-# install updated packages to fix CVE issues
-RUN clean-install zlib1g gzip liblzma5
 
 ENTRYPOINT ["/nfsplugin"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,6 @@ ARG ARCH
 ARG binary=./bin/${ARCH}/nfsplugin
 COPY ${binary} /nfsplugin
 
-RUN apt update && apt upgrade -y && apt-mark unhold libcap2
-RUN clean-install ca-certificates mount nfs-common netbase
+RUN apt update && apt upgrade -y && apt-mark unhold libcap2 && clean-install ca-certificates mount nfs-common netbase
 
 ENTRYPOINT ["/nfsplugin"]


Dockerfile
```diff
@@ -21,6 +21,6 @@ COPY ${binary} /nfsplugin
 RUN apt update && apt-mark unhold libcap2
 RUN clean-install ca-certificates mount nfs-common netbase
 # install updated packages to fix CVE issues
-RUN clean-install libgmp10 bsdutils libssl1.1 openssl libc6 libc-bin libsystemd0 libudev1 zlib1g
+RUN clean-install libgmp10 bsdutils libssl1.1 openssl libc6 libc-bin libsystemd0 libudev1 zlib1g gzip liblzma5
```
pohly, line 24: Wouldn't it make more sense to update *all* packages to the latest version, instead of waiting for a CVE that tells you that an update is needed?

Not all bugs get a CVE. If the distro thinks that a package needs a security update, then I trust that more than the CVE database.

andyzhangx: the cve is reported by 
https://github.com/kubernetes-csi/csi-driver-nfs/blob/700097c7e6127c8758baf0049ab1bb195b35036f/.github/workflows/trivy.yaml#L31-L39

Not sure how to install all packages, there are too many in Ubuntu.
And Ubuntu base image would be upgraded every few months, that may eliminate the necessity to install some packages.

andyzhangx: new image `k8s.gcr.io/build-image/debian-base:bullseye-v1.2.0` is populated, let me adjust the packages a little

andyzhangx: point is you think that's safe, while that image would be scanned by tools like trivy, and trivy would reports CVE error, so it's better to fix it though some CVEs would not be invoked by this image. Anyway, it's better to fix it if it's reported, that's not big effort.

andyzhangx: `debian-base:bullseye-v1.2.0` has more CVE fixes, while it's still missing some fixes, PR updated.

pohly: > Not sure how to install all packages, there are too many in Ubuntu.

`apt update && apt upgrade`?

> Anyway, it's better to fix it if it's reported, that's not big effort.

It's still causing churn in the Dockerfile. Updating all installed packages is safer and never needs to be touched again.

Then when trivy reports the need to update, just tag a new patch release and rebuild the image.

andyzhangx: thanks, `apt upgrade` works!

Dockerfile
```diff
@@ -12,15 +12,13 @@
...
-FROM k8s.gcr.io/build-image/debian-base:bullseye-v1.1.0
+FROM k8s.gcr.io/build-image/debian-base:bullseye-v1.2.0
 
 ARG ARCH
 ARG binary=./bin/${ARCH}/nfsplugin
 COPY ${binary} /nfsplugin
 
-RUN apt update && apt-mark unhold libcap2
+RUN apt update && apt upgrade -y && apt-mark unhold libcap2
 RUN clean-install ca-certificates mount nfs-common netbase
```
pohly, line 22: If you join this into a single RUN, then files that don't need to be in the image get removed before the result of the RUN command is collected:

```suggestion
RUN apt update && apt upgrade -y && apt-mark unhold libcap2 && clean-install ca-certificates mount nfs-common netbase
```

That works because `clean-install` removes the apt cache:
```
# cat /usr/local/bin/clean-install
#!/bin/sh

# Copyright 2021 The Kubernetes Authors.
...
# A script encapsulating a common Dockerimage pattern for installing packages
# and then cleaning up the unnecessary install artifacts.
# e.g. clean-install iptables ebtables conntrack

set -o errexit

if [ $# = 0 ]; then
  echo >&2 "No packages specified"
  exit 1
fi

apt-get update
apt-get install -y --no-install-recommends $@
apt-get clean -y
rm -rf \
   /var/cache/debconf/* \
   /var/lib/apt/lists/* \
   /var/log/* \
   /tmp/* \
   /var/tmp/*
```

andyzhangx: ha, that's quite smart, you are the expert!

pohly (approved):
/lgtm

PR #328, opened 2022-05-04
feat: skip chmod if mountPermissions is 0 after mount

### LINKED ISSUE #327: chmod should not be applied if fsGroupPolicy is set to None
**What happened**:
Set `fsGroupPolicy: None` and mounted a NFSv4 volume that the `fsGroup` could mount but could not modify the mount permissions. Received error caused by: https://github.com/kubernetes-csi/csi-driver-nfs/blob/a42b320c62c1843cef2ee6758c2cbb557a55c6bb/pkg/nfs/nodeserver.go#L117-L119

Per the [Kubernetes CSI Developer Documentation](https://kubernetes-csi.github.io/docs/support-fsgroup.html), `fsGroupPolicy: None` indicates that volumes will be mounted with no modifications, as the CSI volume driver does not support these operations.

Ultimately, the mounting succeeds as the aforementioned line only logs the error.

The CSI driver should not attempt to `chmod` if the `fsGroupPolicy` is set to `None`.

**How to reproduce it**:
Attempt to mount a NFSv4 volume where the `fsGroup` has permission to mount but not modify the volume using `chmod`.

Example use case is where the `fsGroup` only needs to read/write files and does not need to modify permissions to reduce the security risk.

**Environment**:
- CSI Driver version: 3.1.0
- Kubernetes version (use `kubectl version`): 1.23.4
- OS (e.g. from /etc/os-release): fedora-coreos-35.20220410.3.1
- Kernel (e.g. `uname -a`): 5.16.18
- Install tools: Typhoon
- Others:

andyzhangx:
**What type of PR is this?**
/kind feature

feat: skip chmod if mountPermissions is 0 after mount

Fixes #327

```release-note
feat: skip chmod if mountPermissions is 0 after mount
```

### DIFF
--- a/docs/driver-parameters.md
+++ b/docs/driver-parameters.md
@@ -8,7 +8,7 @@ Name | Meaning | Example Value | Mandatory | Default value
 --- | --- | --- | --- | ---
 server | NFS Server address | domain name `nfs-server.default.svc.cluster.local` <br>or IP address `127.0.0.1` | Yes |
 share | NFS share path | `/` | Yes |
-mountPermissions | mounted folder permissions. The default is `0777` |  | No |
+mountPermissions | mounted folder permissions. The default is `0777`, if set as `0`, driver will not perform `chmod` after mount |  | No |
 
 ### PV/PVC usage (static provisioning)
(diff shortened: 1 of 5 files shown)


andyzhangx:
cc @Elegant996

PR #353, opened 2022-06-19
feat: support pv/pvc metadata in subDir parameter

andyzhangx:
**What type of PR is this?**
/kind feature

feat: support pv/pvc metadata in subDir parameter

- `subDir` parameter supports following pv/pvc metadata transform
> if `subDir` value contains following strings, it would be converted into corresponding pv/pvc name or namespace
 - `${pvc.metadata.name}`
 - `${pvc.metadata.namespace}`
 - `${pv.metadata.name}`

Fixes #502

**Requirements**:
- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Release note**:
```
feat: support pv/pvc metadata in subDir parameter
```


andyzhangx:
> /assign @gnufied @jsafrane

@gnufied @jsafrane there is discussion here about the requirement, this nfs driver should have quite similar requirement as smb driver: https://github.com/kubernetes-csi/csi-driver-smb/pull/503#issuecomment-1162687754

PR #370, opened 2022-08-13
feat: disable inline volume by default in kubectl install

### LINKED ISSUE #364: Ephemeral volume lifecycle mode is enabled by default
**What happened**:

The `Ephemeral` volume lifecycle mode is enabled by default in [csi-nfs-driverinfo.yaml](https://github.com/kubernetes-csi/csi-driver-nfs/blob/c78a98288f3d648cf9d6147fa310444a7f228640/deploy/csi-nfs-driverinfo.yaml#L10), which could surprise some admins that deploying this CSI driver allows mounting of any NFS share by users who are only allowed to create pods.

See the Security Considerations section for ephemeral volumes in the [CSI docs](https://kubernetes-csi.github.io/docs/ephemeral-local-volumes.html#security-considerations) and in [KEP-596](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/596-csi-inline-volumes/README.md#security-considerations). The recommendation is:

"CSI inline volumes are only intended to be used for ephemeral storage, and driver vendors should NOT allow usage of inline volumes for persistent storage unless they also provide a third party pod admission plugin to restrict usage of these volumes."

i.e. the `Ephemeral` volume lifecycle mode should either be disabled by default, or a pod admission plugin should be provided with appropriate documentation.

**How to reproduce it**:

**Environment**:
- CSI Driver version: v4.1.0
- Kubernetes version (use `kubectl version`): v1.25.0-alpha
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

andyzhangx:
**What type of PR is this?**
/kind feature

feat: disable inline volume by default in kubectl install

Fixes #364

```release-note
feat: disable inline volume by default in kubectl install
```

### DIFF
--- a/deploy/csi-nfs-driverinfo.yaml
+++ b/deploy/csi-nfs-driverinfo.yaml
@@ -7,5 +7,4 @@ spec:
   attachRequired: false
   volumeLifecycleModes:
     - Persistent
-    - Ephemeral
   fsGroupPolicy: File
--- a/pkg/nfs/controllerserver.go
+++ b/pkg/nfs/controllerserver.go
@@ -364,8 +364,8 @@ func getInternalMountPath(workingMountDir string, vol *nfsVolume) string {
 
 // Get internal path where the volume is created
 // The reason why the internal path is "workingDir/subDir/subDir" is because:
-//   * the semantic is actually "workingDir/volId/subDir" and volId == subDir.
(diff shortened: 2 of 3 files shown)


dobsonj:
/lgtm
thanks!

PR #375, opened 2022-09-06
feat: parametrization of deployment strategy for csi controller

mariouhrin:
**What type of PR is this?**
/kind feature

It allows the options to override default deployment strategy for csi controller (from RollingUpdate to Recreate).

In my company, we are using single node k3s cluster. We had an issue with already allocated port after new deployments of csi-driver-nfs helm chart.

This is the error logs from csi controller pod:
Warning  FailedScheduling  12m (x42 over 44m)   default-scheduler  0/1 nodes are available: 1 node(s) didn't have free ports for the requested pod ports.

The problematic port was this one (csi controller deployment manifest):
```yaml
- containerPort: 29652                                                                                                                                                                                                                              
  hostPort: 29652                                                                                                                                                                                                                                            
  name: healthz                                                                                                                                                                                                                                     
  protocol: TCP
```
 
The Recreate deployment strategy helped us to avoid this already allocated port issue. The RollingUpdate deployment strategy is probably causing a problem when the old and new pods are running at the same time and 29652 port cannot be used for new pod. But maybe I am wrong.

Fixes #

```release-note
feat: parametrization of deployment strategy for csi controller
```


charts/latest/csi-driver-nfs/values.yaml
```diff
@@ -38,6 +38,7 @@ kubeletDir: /var/lib/kubelet
 controller:
   name: csi-nfs-controller
   replicas: 1
+  strategyType: RollingUpdate
```
andyzhangx, line 41: Thanks, I think we could change the default value as `Recreate`, that could avoid driver install issue when reinstall driver on a cluster with only a few nodes.

mariouhrin: Alright, I'm going to change it

andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm
/approve

Issue #459, opened 2023-06-04
VolumeSnapshot path


saif-88:
**What happened**:

When i create a VolumeSnapshot the path in the nfs server contain the full path of the PersistentVolume
PersistentVolume path : /mnt/nfs/pvc-xxxxxx
VolumeSnapshot path /mnt/nfs/mnt/nfs/snapcontent-xxxxxx

**What you expected to happen**:

VolumeSnapshot path /mnt/nfs/snapcontent-xxxxxx

**How to reproduce it**:
```yaml
---
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshotClass
metadata:
  name: snapshot-csi
  annotations:
    snapshot.storage.kubernetes.io/is-default-class: "true"
driver: nfs.csi.k8s.io
deletionPolicy: Delete
parameters:
  server: xxxxxx
  share: "/mnt/nfs"
---
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: snapshot-test
spec:
  source:
    persistentVolumeClaimName: pvc-test
```

**Anything else we need to know?**:

**Environment**:

- CSI Driver version: 4.3.0
- Kubernetes version (use `kubectl version`): 1.26.5 (v1.26.5+rke2r1)
- OS (e.g. from /etc/os-release): Debian 11.5
- Kernel (e.g. `uname -a`): 5.10.0-19-amd64
- Install tools:
- Others:

andyzhangx:
@saif-88 is there sth. broken with this snapshot feature? btw, following parameters are not supported in snapshot class:
```
parameters:
  server: xxxxxx
  share: "/mnt/nfs"
```

saif-88:
@andyzhangx Thank you for response
the path in the nfs server should be 
/mnt/nfs/snapcontent-xxxxxx
and not 
/mnt/nfs/mnt/nfs/snapcontent-xxxxxx

For the parameters part, i saw that in the code of the controller and CreateSnapshotRequest spec and the snapshots are placed in the directory set in the share parameter when i change it

wozniakjan:
I suppose you have a `StorageClass` with `  share: "/mnt/nfs"` as well, it looks like I messed up some part of the https://github.com/kubernetes-csi/csi-driver-nfs/pull/430 and somehow it maybe concatenates `StorageClass` share with `VolumeSnapshotClass` share? I will take a look and test this more thoroughly. 

[...]

PR #430, opened 2023-03-18
feat: implement snapshots as tarballs

wozniakjan:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
This PR proposes simple support for `VolumeSnapshots` as tarballs created with 
```
tar -C /[src_vol_path] -czvf /[snapshot_path]/[snapshot_name]/[src_vol_name].tar.gz .
```

The reason why `[src_vol_uuid]` is included in the snapshot path is to be able to efficiently discover [`AlreadyExists` cases](https://github.com/kubernetes-csi/csi-test/blob/3fd3e7915d1029e10e81923e4e0cb3de23059c20/pkg/sanity/controller.go#L1470)

It is the second out of three intended PRs to provide basic snapshot workflow including volume cloning from snapshots. It is preceded by https://github.com/kubernetes-csi/csi-driver-nfs/pull/426 which implemented volume cloning from other volumes.

**Which issue(s) this PR fixes**:

Moving forward with snapshots as discussed in https://github.com/kubernetes-csi/csi-driver-nfs/issues/31

**Special notes for your reviewer**:
Some parts of this PR related to helm chart and external-snapshotter are not ideal. As a followup task, I would like to attempt to address https://github.com/kubernetes-csi/external-snapshotter/issues/812. That way all csi-driver helm charts including this one, instead of embedding external-snapshotter CRDs and controller resources and duplicating them, they could just add the official external-snapshotter chart as a dependency. This should also simplify updating to new releases. But not in scope to be addressed here.

**Does this PR introduce a user-facing change?**:

```release-note
feat: volume snapshots
```

PR #461, opened 2023-06-05
bugfix: correct snapshot path when using share parameter

wozniakjan:
**What type of PR is this?**
/kind bug

When using the `share` parameter on either `StorageClass` or `SnapshotVolumeClass`, the path to snapshots would contain the `share` parameter twice, e.g.
```
/mnt/nfs/mnt/nfs/snapshot-id/src-pvc-id.tar.gz   # duplicate
/mnt/nfs/snapshot-id/src-pvc-id.tar.gz           # desired
```
This PR ensures it's the desired path.

Fixes https://github.com/kubernetes-csi/csi-driver-nfs/issues/459

Should I also introduce some special treatment regarding the old broken snapshot paths? For those who already have created some snapshots and they are stored in the broken path.

```release-note
bugfix: correct snapshot path when using share parameter
```

### DIFF
--- a/pkg/nfs/controllerserver_test.go
+++ b/pkg/nfs/controllerserver_test.go
@@ -742,10 +742,10 @@ func TestCopyVolume(t *testing.T) {
 				uuid:    "dst-pv-name",
 			},
 			prepare: func() error {
-				if err := os.MkdirAll("/tmp/snapshot-name/share/snapshot-name/", 0777); err != nil {
+				if err := os.MkdirAll("/tmp/snapshot-name/snapshot-name", 0777); err != nil {
 					return err
 				}
-				file, err := os.Create("/tmp/snapshot-name/share/snapshot-name/src-pv-name.tar.gz")
+				file, err := os.Create("/tmp/snapshot-name/snapshot-name/src-pv-name.tar.gz")
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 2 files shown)


andyzhangx:
let's fix this issue in master branch first, and for the existing issue, we could add a guide asking user to migrate.

wozniakjan:
ack, in that case, the PR should be addressing the bug and is ready for review, thanks @andyzhangx!

I will follow-up with a migration guide, I think what @saif-88 described in https://github.com/kubernetes-csi/csi-driver-nfs/issues/459#issuecomment-1576991260 is feasible.

andyzhangx (approved):
/lgtm

pkg/nfs/controllerserver_test.go
```diff
@@ -742,10 +742,10 @@ func TestCopyVolume(t *testing.T) {
 				uuid:    "dst-pv-name",
 			},
 			prepare: func() error {
-				if err := os.MkdirAll("/tmp/snapshot-name/share/snapshot-name/", 0777); err != nil {
+				if err := os.MkdirAll("/tmp/snapshot-name/snapshot-name", 0777); err != nil {
```
wozniakjan, line 745: I should have been more careful when writing unit tests, this case was clearly showing the bug reported by https://github.com/kubernetes-csi/csi-driver-nfs/issues/459 and I just didn't realize it during https://github.com/kubernetes-csi/csi-driver-nfs/pull/430

PR #462, opened 2023-06-07
chore: forbid invalid parameters in snapshot storage class

andyzhangx:
**What type of PR is this?**
/kind bug

chore: forbid invalid parameters in snapshot storage class

Fixes #

```release-note
none
```

### DIFF
--- a/pkg/nfs/controllerserver.go
+++ b/pkg/nfs/controllerserver.go
@@ -568,6 +568,8 @@ func newNFSSnapshot(name string, params map[string]string, vol *nfsVolume) (*nfs
 			server = v
 		case paramShare:
 			baseDir = v
+		default:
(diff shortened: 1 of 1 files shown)


wozniakjan (approved):
/lgtm

but I think it will not respect the error type
https://github.com/kubernetes-csi/csi-driver-nfs/blob/8cee28152d4c990d6f1add72ff0b1c13977dd3bc/pkg/nfs/controllerserver.go#L313-L315

nit: is it desirable to also have a similar check for `StorageClass` parameters?
https://github.com/kubernetes-csi/csi-driver-nfs/blob/8cee28152d4c990d6f1add72ff0b1c13977dd3bc/pkg/nfs/controllerserver.go#L603-L620

PR #481, opened 2023-07-13
feat: allow csi-snapshotter to be enabled without installing CRDs

### LINKED ISSUE #467: Allow csi-snapshotter to be enabled without installing CRDs
Currently, to enable the `csi-snapshotter` container in the controller Deployment, you must set `.Values.externalSnapshotter.enabled=true`.  This also causes the CRDs and snapshot-controller to be installed.  This causes a conflict if the underlying kubernetes distro also installs the CRDs and snapshot-controller

Either fully decouple the `csi-snapshotter` container from the installation of CRDs/snapshot-controller, or add an override value so this can be enabled without adding CRDs/snapshot-controller

See 2 options above.

navilg:
**What type of PR is this?**

/kind bug

When VolumeSnapshot,VolumeSnapshotClass or VolumeSnapshotContent CRD already exists on K8s cluster, helm fails to install the chart on cluster if externalSnapshotter is enabled.

With these changes, externalSnapshotter can be enabled without trying to re-create those CRDs if they are already present in cluster.

Fixes #467

```release-note
To deploy csi-snapshotter without installing any of CRDs, Install with below flags:

--set externalSnapshotter.customResourceDefinitions.enabled=false
```


gclawes:
Is there ever a situation where only some of the snapshot CRD should be enabled?

wozniakjan (commented):
> Is there ever a situation where only some of the snapshot CRD should be enabled?

there are imho two scenarios
1) the user has external-snapshotter already installed for some other CSI driver
 - in this case, it makes sense to not install CRDs and resources related to the [snapshot-controller](https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/deploy/csi-snapshot-controller.yaml)
2) the user doesn't have external-snapshotter at all
 - in this case, they will need all of the CRDs and the controller as well

navilg:
> I would be in favour of introducing just a single config knob for the snahpshot-controller related resources and CRDs combined, wdyt?

A cluster may have multiple csi-drivers of other third-party or cloud-native storage solutions. In those cases (scenario 1), CRDs will already exist and helm install will fail. So single config knob for snapshot-controller and CRDs combined would be problematic here. 

imo, if external-snapshotter is enabled AND CRDs are enabled, in that case only CRDs should be deployed. Instead of multiple knobs (each of one CRDs), We may look into having single knob for all 3 CRDs. And we can keep external snapshotter and CRDs enabled by default.

gclawes:
> Is there ever a situation where only some of the snapshot CRD should be enabled?

My point is more why do we need a separate boolean value for each of the 3 CRDs?  Why not a single one to enable/disable all 3 CRDs.  I don't think there's ever a situation where a cluster admin would only install 2 of the 3 CRDs for snapshots, right?


Also, I'm getting a parse error with the latest version of the PR:
```
$ helm upgrade --install csi-driver-nfs ~/repos/github.com/navilg/csi-driver-nfs/charts/latest/csi-driver-nfs --namespace kube-system --version v0.0.0 -f helm/csi-driver-nfs-values.yaml
Release "csi-driver-nfs" does not exist. Installing it now.
Error: parse error at (csi-driver-nfs/templates/crd-csi-snapshot.yaml:2): "-"
```

for values:
```
externalSnapshotter:
  customResourceDefinitions:
    volumeSnapshot: false
    volumeSnapshotClass: false
    volumeSnapshotContent: false
```

wozniakjan:
> My point is more why do we need a separate boolean value for each of the 3 CRDs?

+1

> But the original issue in #467 (clarified here) was the need to decouple:
> * The installation of the CRDs and the snapshot-controller Deployment
> * The csi-snapshotter sidecar in the csi-nfs-controller Deployment.

+1 and all of my thumbs up

navilg:
> My point is more why do we need a separate boolean value for each of the 3 CRDs? Why not a single one to enable/disable all 3 CRDs. I don't think there's ever a situation where a cluster admin would only install 2 of the 3 CRDs for snapshots, right?

Yeah. We can have one config to enable/disable all 3 CRDs instead of having them individually enabled/disabled.

navilg:
> * The `csi-snapshotter` sidecar in the `csi-nfs-controller` `Deployment`.

So a separate knob for CRDS (one for all  three CRDs) and another knob for `csi-snapshotter` sidecar. I think we can add this. Let me know if I should go ahead and try this @wozniakjan @gclawes . Or do you have any other alternatives

wozniakjan:
> So a separate knob for CRDS (one for all three CRDs) and another knob for `csi-snapshotter` sidecar

I would keep the sidecar controlled only by [`.externalSnapshotter.enabled`](https://github.com/kubernetes-csi/csi-driver-nfs/blob/748bde32e0f8cf23d41260cef14fed1dd07d7804/charts/v4.4.0/csi-driver-nfs/values.yaml#L128). But please consider coupling the [`snapshot-controller`](https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/charts/v4.4.0/csi-driver-nfs/templates/csi-snapshot-controller.yaml) with the external-snapshotter CRDs. If a cluster has these CRDs for another CSI driver, it likely has the controller as well.

gclawes:
I think this works.  Another option is to use the `crds/` folder of helm3: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you

That way, CRDs can be skipped with the `helm install --skip-crds` flag.   Using this approach means updates to the CRDs must be applied manually though, as cautioned in the helm docs.

There is really no "good" way to do CRDs with helm, only least-bad ways depending on the situation (for example`kube-prometheus-stack` discussions below):
https://github.com/prometheus-community/helm-charts/issues/2921
https://github.com/prometheus-community/helm-charts/issues/2612

kfox1111:
If moved to the crds dir, make sure you have at least one release where the old crd has the helm annotation to cause it to not delete it when its removed from helm manifest maintenance.

gclawes:
> If moved to the crds dir, make sure you have at least one release where the old crd has the helm annotation to cause it to not delete it when its removed from helm manifest maintenance.

Yeah, I think this would require very specific upgrade instructions to preserve CRD state without accidentally removing them.

andyzhangx:
> > My point is more why do we need a separate boolean value for each of the 3 CRDs?
> 
> +1
> 
> > But the original issue in #467 (clarified here) was the need to decouple:
> > 
> > * The installation of the CRDs and the snapshot-controller Deployment
> > * The csi-snapshotter sidecar in the csi-nfs-controller Deployment.
> 
> +1 and all of my thumbs are up

agree, we only need on boolean for all the 3 CRDs

navilg:
Replaced three boolean values with single boolean value to install (or not install) all 3 CRDs.

andyzhangx:
@navilg could you rebase to master branch and also squash all commits? thanks.

charts/latest/csi-driver-nfs/templates/crd-csi-snapshot.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.externalSnapshotter.enabled -}}
+{{- if and .Values.externalSnapshotter.enabled .Values.externalSnapshotter.customResourceDefinitions.enabled -}}
```
kfox1111, line 1: Cant comment on line 8 below, but need (annotation) added:
 "helm.sh/resource-policy": keep

navilg: Yup. We can add it

kfox1111:
Not sure I need to file another issue for this or not....

Say I have the external snapshoter installed via: https://artifacthub.io/packages/helm/piraeus-charts/snapshot-controller

With this flag it still seems like there is some missing functionality to be able to share a snapshot-controller like this?

https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/charts/v4.4.0/csi-driver-nfs/templates/csi-nfs-controller.yaml#L69-L85 is still needed to be enabled, but the whole controller is also deployed in that case? https://github.com/kubernetes-csi/csi-driver-nfs/blob/master/charts/v4.4.0/csi-driver-nfs/templates/csi-snapshot-controller.yaml#L1

andyzhangx (approved):
/lgtm

andyzhangx:
@kfox1111 would fix by https://github.com/kubernetes-csi/csi-driver-nfs/pull/490

PR #490, opened 2023-07-20
fix: always enable snapshot sidecar container in controller

andyzhangx:
**What type of PR is this?**
/kind bug

fix: always enable snapshot sidecar container in controller

when external snapshot controller is disabled, this snapshot sidecar container should also be enabled by default.
this PR is to address https://github.com/kubernetes-csi/csi-driver-nfs/pull/481#issuecomment-1644044642

Fixes #

```release-note
fix: always enable snapshot sidecar container in controller
```


gclawes:
Will this be in the next release of the driver?

andyzhangx:
> Will this be in the next release of the driver?

@gclawes since this is just helm chart change, will be fixed by https://github.com/kubernetes-csi/csi-driver-nfs/pull/495

PR #528, opened 2023-10-23
go.mod: Fix unresolvable indirect dependency

pierreprinetti:
Fix `go list -mod=readonly -m all` by adding a replace for `k8s.io/dynamic-resource-allocation` that points to the Kubernetes version currently in use.

The packages is not used here, but making the dependency resolvable makes the go toolchain happy.

cf.: https://github.com/kubernetes/kubernetes/blob/4b48ab1fdbdeee3568790e44bea55ea719205025/go.mod#L265

```release-note
none
```


MaysaMacedo:
LGTM

PR #603, opened 2024-02-05
charts: fix master and control-plane nodeSelectors

intelfx:
**What type of PR is this?**
/kind bug

`controller.runOnMaster` and `controller.runOnControlPlane` are non-functional as written, since an empty value in a nodeSelector is not treated as a wildcard. Fix the nodeSelectors to match against `"true"`, which is enough to successfully schedule the controller on a master/control-plane node in my testing.

```release-note
none
```

### DIFF
--- a/charts/latest/csi-driver-nfs/templates/csi-nfs-controller.yaml
+++ b/charts/latest/csi-driver-nfs/templates/csi-nfs-controller.yaml
@@ -31,10 +31,10 @@ spec:
       nodeSelector:
         kubernetes.io/os: linux
         {{- if .Values.controller.runOnMaster}}
-        node-role.kubernetes.io/master: ""
+        node-role.kubernetes.io/master: "true"
         {{- end}}
         {{- if .Values.controller.runOnControlPlane}}
-        node-role.kubernetes.io/control-plane: ""
+        node-role.kubernetes.io/control-plane: "true"
         {{- end}}
 {{- with .Values.controller.nodeSelector }}
 {{ toYaml . | indent 8 }}


charts/latest/csi-driver-nfs/templates/csi-nfs-controller.yaml
```diff
@@ -31,10 +31,10 @@ spec:
       nodeSelector:
         kubernetes.io/os: linux
         {{- if .Values.controller.runOnMaster}}
-        node-role.kubernetes.io/master: ""
+        node-role.kubernetes.io/master: "true"
         {{- end}}
         {{- if .Values.controller.runOnControlPlane}}
-        node-role.kubernetes.io/control-plane: ""
+        node-role.kubernetes.io/control-plane: "true"
```
andyzhangx, line 37: not sure whether all master node are labeled with `node-role.kubernetes.io/control-plane: ""`, but as I know, the azure capz cluster are labeled as `node-role.kubernetes.io/control-plane: ""`

andyzhangx: I think you could set `set --nodeSelector=xxx` to make it work on your testing env if existing config does not work on your testing env.

intelfx: I can, but it's a footgun. Also, as I remember, at least GKE also has `"true"` instead of `""`.

Would you be open to a patch that replaces these node selector with node affinities (which make it possible to declare an "Exists" relation instead of a strict match)?

PR #635, opened 2024-03-22
setup a variable on values to allow specify a repo URL

floreseo2:
**What type of PR is this?**
/kind feature

In airgapped enviroment you need to setup a local repo where you upload images. In order to easy configure this local repo "repourl" variable has been created. With the default value user seems the same behaivour 

Fixes #

```release-note
none
```


charts/v4.6.0/csi-driver-nfs/templates/csi-nfs-controller.yaml
```diff
@@ -98,7 +98,7 @@ spec:
           securityContext:
             readOnlyRootFilesystem: true
         - name: nfs
-          image: "{{ .Values.image.nfs.repository }}:{{ .Values.image.nfs.tag }}"
+          image: "{{ .Values.image.repourl }}{{ .Values.image.nfs.repository }}:{{ .Values.image.nfs.tag }}"
```
andyzhangx, line 101: this may break existing chart users since only nfs.repository is set in old versions, what about write like following?

```
        - name: nfs
{{- if hasPrefix "/" .Values.image.nfs.repository }}
          image: "{{ .Values.image.baseRepo }}{{ .Values.image.nfs.repository }}:{{ .Values.image.nfs.tag }}"
{{- else }}
          image: "{{ .Values.image.nfs.repository }}:{{ .Values.image.nfs.tag }}"
{{- end }}
```

andyzhangx (commented):
pls also sign the cla, thanks.

charts/v4.6.0/csi-driver-nfs/templates/csi-nfs-controller.yaml
andyzhangx: example: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/fda1d0a3174ce628d57ce01a86b15a492a77f583/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml#L152C1-L157C11

floreseo2:
Close to follow suggestion done by  @andyzhangx

andyzhangx:
would be addressed by https://github.com/kubernetes-csi/csi-driver-nfs/pull/637

PR #639, opened 2024-03-28
feat: use new restricted liveness probe endpoint

anothertobi:
**What type of PR is this?**

/kind feature

The `--health-port` option is deprecated and `--http-endpoint` the new
option to use. Since `--http-endpoint` additionally exposes metrics, the
liveness probe is limited to only listen on localhost by default. With
this only processes on the host network, like kubelet, can call the
endpoint. Regular pods can no longer access the liveness probe via the
node IP address.

n/a

n/a

```release-note
Updated the liveness probe endpoints to only listen on localhost by default.
```


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm

PR #720, opened 2024-07-23
fix: tmp path mount permission error on charts

andyzhangx:
**What type of PR is this?**
/kind bug

fix: tmp path mount permission error

Fixes #

```release-note
fix: tmp path mount permission error
```


rhessus:
Hi,

Since this issue was delivered, I had to apply a Kustomize patch to restore the tmp-dir volume, which was removed and caused the following error:

`failed to provision volume with StorageClass "nfs-csi-xxx-sc": rpc error: code = Internal desc = failed to mount nfs server: rpc error: code = Internal desc = mkdir /tmp/pvc-6a531286-d87b-4e11-95cd-39cfa7a9d6c0: read-only file system`

Here is the patch Kustomize i applied to work around the issue:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: csi-nfs-controller
  namespace: kube-system
spec:
  template:
    spec:
      containers:
        - name: nfs
          volumeMounts:
          - mountPath: /tmp
            name: tmp-dir
      volumes:
        - emptyDir: {}
          name: tmp-dir
```

andyzhangx:
> Hi,
> 
> Since this issue was delivered (since v4.8.0 of the chart), I had to apply a Kustomize patch to restore the tmp-dir volume, which was removed and caused the following error:
> 
> `failed to provision volume with StorageClass "nfs-csi-xxx-sc": rpc error: code = Internal desc = failed to mount nfs server: rpc error: code = Internal desc = mkdir /tmp/pvc-6a531286-d87b-4e11-95cd-39cfa7a9d6c0: read-only file system`
> 
> Here is the patch Kustomize i applied to work around the issue:
> 
> ```yaml
> apiVersion: apps/v1
> kind: Deployment
> metadata:
>   name: csi-nfs-controller
>   namespace: kube-system
> spec:
>   template:
>     spec:
>       containers:
>         - name: nfs
>           volumeMounts:
>           - mountPath: /tmp
>             name: tmp-dir
>       volumes:
>         - emptyDir: {}
>           name: tmp-dir
> ```
> 
> Thanks to this patch, I was finally able to upgrade the chart from version 4.7.0 to 4.11.0.
> 
> For context, I am running this within a MicroK8s cluster on three Raspberry Pi 5 nodes.
> 
> Could you confirm if the removal of tmp-dir was intentional? If so, what would be the recommended way to handle this situation?

@rhessus yes, removal of tmp-dir is intentional, default workingMountDir is `/tmp` on the node now, you could customize this setting to use other place.

https://github.com/kubernetes-csi/csi-driver-nfs/blob/3e189521d06f0b3daf10fddb806665de3fa7688a/charts/latest/csi-driver-nfs/values.yaml#L64

PR #830, opened 2024-12-20
do not generate snapshotter container if enabled: false

alexgaganashvili:
**What type of PR is this?**

/kind cleanup

Do not include snapshotter container unless .externalSnapshotter.enabled is true

```release-note
none
```

### DIFF
--- a/charts/v4.9.0/csi-driver-nfs/templates/csi-nfs-controller.yaml
+++ b/charts/v4.9.0/csi-driver-nfs/templates/csi-nfs-controller.yaml
@@ -86,6 +86,7 @@ spec:
             capabilities:
               drop:
               - ALL
+{{- if .Values.externalSnapshotter.enabled }}
         - name: csi-snapshotter
 {{- if hasPrefix "/" .Values.image.csiSnapshotter.repository }}
           image: "{{ .Values.image.baseRepo }}{{ .Values.image.csiSnapshotter.repository }}:{{ .Values.image.csiSnapshotter.tag }}"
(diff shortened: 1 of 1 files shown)


charts/v4.9.0/csi-driver-nfs/templates/csi-nfs-controller.yaml
```diff
@@ -86,6 +86,7 @@ spec:
             capabilities:
               drop:
               - ALL
+{{- if .Values.externalSnapshotter.enabled }}
```
andyzhangx, line 89: we have already `controller.enableSnapshotter` to control this, and even snapshot controller is disabled, some users still need this snapshotter container to run there:

https://github.com/kubernetes-csi/csi-driver-nfs/blob/49570c776643a8bd6a86784456a150013ef2dccd/charts/latest/csi-driver-nfs/values.yaml#L59

alexgaganashvili: If I don't need it, I can just sent the resources to minimum, I guess? How do I know if I need the container to run, since I do not need to snapshot controller?

andyzhangx: if you don't want snapshot functionality, then you could set `controller.enableSnapshotter` as false

alexgaganashvili: But the snapshotter container will still be included. I'm not sure why I need it there by default.

andyzhangx: the change is made in https://github.com/kubernetes-csi/csi-driver-nfs/pull/820/files, in next released version, you could use `--set controller.enableSnapshotter=false` to disable snapshotter container

PR #901, opened 2025-04-20
feat: add POD_NAMESPACE environment var

mfhunruh:
**What type of PR is this?**
/kind feature

It's better to use a dynamic environment variable `POD_NAMESPACE` based on the value of the `metadata.namespace` field. This will simplify deployment for those who use kubectl/kustomize instead of Helm.

Fixes #

```release-note
none
```


andyzhangx (approved):
/ok-to-test
/lgtm

PR #999, opened 2025-11-05
feat: authentication with Kerberos credentials

### LINKED ISSUE #525: NFS credentials
Right now it seems that this driver can only mount NFS volumes if server allow anonymous access.
2 years ago, on https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner/issues/136#issuecomment-907710848 was said that this feature could have been implemented here  but I wasn't able to find anything related.

A way to set some kind of credentials to perform authentication (eg Kerberos, sys) 

**Additional context**
none

GamerGirlandCo:
**What type of PR is this?**

/kind feature

This PR implements authentication with a kerberos principal and password when mounting NFS shares.

Fixes #525

- **I changed the Dockerfile's base image to `debian:stable-slim`**
  Justification: services cannot be started due to a missing file `/lib/lsb/init-functions`
- **I changed the Dockerfile's entrypoint to a script (`entry.sh`)**
  The script is created in the Dockerfile inline, and you can review it [here](https://github.com/GamerGirlandCo/csi-driver-nfs/blob/21956ed61da89132c94d1192a00fa25562fc2c6b/Dockerfile#L32-L41)
  Justification: this is so we can start necessary services (`rpcbind` and `nfs-common`) if and only if we're in the controller pod. Otherwise, the `nfsplugin` binary is invoked with the exact same arguments the pod's `nfs` container was started with.

```release-note
Added the ability to authenticate with kerberos-protected NFS shares
The following new parameters may be specified when creating or updating a storage class:
- **`authPrincipal`**
  The principal that will be used to authenticate to the share
- **`authKeytab`**
  The secret containing base64-encoded keytab content
- **`authKrbConf`**
  The secret containing contents of a krb5.conf with information about how to connect to a realm.
```

---

## Investigation update (2026-08-18) — Kerberos e2e mount timeout

CI job `pull-csi-driver-nfs-e2e-capz` run [`2089605852326531072`](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-csi_csi-driver-nfs/999/pull-csi-driver-nfs-e2e-capz/2089605852326531072) — the Kerberos mount `mount.nfs4 -o sec=krb5,vers=4.1 nfs-krb-server.default.svc.cluster.local:/srv/shared` hangs 110s and fails with `mount.nfs4: Connection timed out`.

### Debug artifacts pulled
- Client side (controller pod): [`csi-nfs-controller-*/nfs.log`](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-csi_csi-driver-nfs/999/pull-csi-driver-nfs-e2e-capz/2089605852326531072/artifacts/clusters/capz-9j4atl/kube-system/) — `rpc.gssd -vf` shows two successful `do_downcall` with 24h lifetime, so client-side GSS init is fine.
- Server side (`nfs-krb-server-7cd8697649-lkjxj/nfs-krb-server.log`) — KDC + mountd + tail of `/var/log/rpc-gssd.log`; **`/var/log/gssd.log` (rpc.svcgssd verbose log) is empty for the whole test window**.

### Timeline (server-side, 07:18–07:20 UTC)
- `07:18:57` KDC `AS_REQ ISSUE` — kinit for `nfs/nfs-krb-server...` (password auth)
- `07:18:58` `AS_REQ ISSUE` — kinit `-k` from keytab
- `07:19:02` `AS_REQ ISSUE` — rpc.gssd machine credential
- `07:19:03` `TGS_REQ ISSUE` — `nfs/nfs-krb-server...` requests service ticket for **itself** (only principal in the KDC)
- `07:19:03` `rpc.mountd: successful authentication for IP 192.168.163.0 as *` → wildcard export accepted
- **Then silence for 90+ seconds until `mount.nfs4` gives up.**
- No entry in `rpc-gssd.log` / `gssd.log` from svcgssd → **kernel nfsd never up-called svcgssd for context verification**.

### Root-cause candidates on the server image (`thealmightydrawingtablet/nfs-krb:alpine`)
Extracted `entry.sh`/`init.sh` from the image confirm:

1. **KDC only ever provisions one principal**: `nfs/nfs-krb-server.default.svc.cluster.local`. There is no separate `host/…` or client user principal. The client's `rpc.gssd` reuses this principal as a machine cred → server receives context where `client == server`, `svcgssd` silently drops it.
2. **`sec=krb5p:krb5i:krb5` in exports with `krb5p` first**: negotiation may fail against the client mount opt `sec=krb5` (integrity only) on some kernels; there is no `sec=sys` fallback for v4.1 SEQUENCE.
3. **idmapd `Domain` is set to the realm string** `NFS-KRB-SERVER.DEFAULT.SVC.CLUSTER.LOCAL` (not a DNS-style domain) — this is known to stall NFSv4 `PUTROOTFH`/`GETATTR` when nfsd tries to id-map the caller.
4. `entry.sh` restarts `rpc.svcgssd` manually after start (`pidof rpc.svcgssd` loop), suggesting the packaged init already had issues.
5. Sanity: from the **same** controller pod a plain `sec=sys` mount of an unrelated `nfs-server` succeeded, so networking / DNS / RPC portmap are all healthy — problem is 100% inside the krb5 code path.

### What is NOT the problem (previously ruled out)
- `noresvport`: controller runs with `hostNetwork+privileged+SYS_ADMIN`, server exports have `insecure`. Removing `noresvport` did not change the outcome.
- Client-side GSS init: `do_downcall` succeeds twice with normal lifetime.
- Plain NFS mount from the same client pod: works, so no CNI/DNS/portmap regression.

### Proposal
The image / yaml under `test/nfs-krb-server/` needs at least (any one of these is likely enough):
- Provision a **separate host/machine principal** for the client and mount its keytab into the controller pod, OR
- Simplify export sec list to just `sec=krb5` and add `sec=sys` fallback so v4.1 session bring-up doesn't hinge on GSS, OR
- Fix `/etc/idmapd.conf` `Domain=` to a DNS name and align `Local-Realms`.

Until the server image is fixed, we can:
- **`ginkgo.Skip`** the Kerberos e2e case (label `[Kerberos]`) so this PR can proceed on its non-Kerberos merits (rpc.gssd lifecycle, secret plumbing, mountPermissions), OR
- Split the server fix into a follow-up PR and gate the e2e on that image tag.

/cc @andyzhangx

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,13 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.k8s.io/build-image/debian-base:bookworm-v1.0.6
+FROM debian:stable-slim
 
 ARG ARCH
 ARG binary=./bin/${ARCH}/nfsplugin
 COPY ${binary} /nfsplugin
 
-RUN apt update && apt upgrade -y && apt-mark unhold libcap2 && clean-install ca-certificates mount nfs-common netbase krb5-user
+RUN apt update && apt upgrade -y && apt-mark unhold libcap2 && apt-get install -y --reinstall --purge ca-certificates mount nfs-common netbase krb5-user lsb-base bash
 
 RUN cat > /etc/default/nfs-common <<EOC
 NEED_STATD=yes
@@ -28,4 +28,18 @@ NEED_IDMAPD=yes
 NEED_GSSD=yes
 EOC
 
-ENTRYPOINT ["/nfsplugin"]
+RUN cat > /usr/local/bin/entry.sh <<'EOF'
+#!/bin/sh
+set -x
+
+if [ "$1" = "true" ]; then
+	shift 1
+	service rpcbind start
+	service nfs-common start
+fi
+
+/nfsplugin $@
+EOF
+RUN chmod +x /usr/local/bin/entry.sh
+
+ENTRYPOINT ["entry.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,7 @@ if [ "$1" = "true" ]; then
 	shift 1
 	service rpcbind start
 	service nfs-common start
+	sleep 5
 fi
 
 /nfsplugin $@
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,14 @@ ARG ARCH
 ARG binary=./bin/${ARCH}/nfsplugin
 COPY ${binary} /nfsplugin
 
-RUN apt update && apt upgrade -y && apt-mark unhold libcap2 && clean-install ca-certificates mount nfs-common netbase
+RUN apt update && apt upgrade -y && apt-mark unhold libcap2 && clean-install ca-certificates mount nfs-common netbase krb5-user
+
+RUN cat > /etc/default/nfs-common <<EOC
+NEED_STATD=yes
+
+NEED_IDMAPD=yes
+
+NEED_GSSD=yes
+EOC
 
 ENTRYPOINT ["/nfsplugin"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,10 +34,11 @@ set -x
 
 if [ "$1" = "true" ]; then
 	shift 1
+fi
 
 	# Debian 13 sysvinit scripts assume systemd-created runtime dirs.
 	# Create the paths they touch so `service rpcbind start` doesn't fail.
-	mkdir -p /run/sendsigs.omit.d /var/lib/nfs/rpc_pipefs /var/lib/nfs/statd /var/lib/nfs/v4recovery /run/rpc_pipefs
+	mkdir -p /run/sendsigs.omit.d /var/lib/nfs/statd /var/lib/nfs/v4recovery /run/rpc_pipefs
 
 	# rpc.gssd needs rpc_pipefs mounted; nfs-common sysvinit will happily
 	# report "gssd." as started even when it isn't, causing sec=krb5 mounts
@@ -55,7 +56,6 @@ if [ "$1" = "true" ]; then
 	fi
 
 	sleep 5
-fi
 
 /nfsplugin $@
 EOF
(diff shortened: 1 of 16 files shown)


andyzhangx (commented):
/ok-to-test

andyzhangx:
The e2e failure on `should create a volume with kerberos auth` is caused by the Debian 13 base image's incomplete sysvinit compat:

- `rpcbind` init script fails because `/run/sendsigs.omit.d/` doesn't exist
- `nfs-common` init script prints `gssd.` as if it started but `rpc.gssd` never actually runs (no `rpc_pipefs` mount), so any `sec=krb5` mount hangs 110s waiting on GSS context

Non-krb NFSv4 mounts still work because they don't need `rpc.gssd`, which is why only that one test times out.

Pushed c50834e to fix entry.sh:
- Pre-create the runtime dirs the init scripts expect
- Explicitly `mount -t rpc_pipefs sunrpc /var/lib/nfs/rpc_pipefs` before starting nfs-common
- Fall back to launching `rpc.gssd` directly if sysvinit didn't
- Add `procps` for `pgrep`

Container is already privileged + `SYS_ADMIN` so the rpc_pipefs mount is permitted.

andyzhangx:
Second CI run (2089174931572002816) shows progress from c50834e:
- `service rpcbind start` now succeeds (`Starting RPC port mapper daemon: rpcbind.`)
- `pgrep -x rpc.gssd` finds the process (fallback launch not needed)

But the kerberos mount still times out at 110s. Almost certainly because **rpc.gssd was started at container init, before `/etc/krb5.conf` and `/etc/krb5.keytab` existed** — those files are written per-mount inside `NodePublishVolume`, and rpc.gssd caches its GSSAPI config at startup and never re-reads them. So the kernel GSS upcall never resolves and mount(2) blocks.

Pushed 85726aa:
- After kinit succeeds and before the mount, `pkill rpc.gssd` and re-`exec /usr/sbin/rpc.gssd -vvv` so it picks up the fresh krb5.conf/keytab
- Capture ktutil / kinit stdout+stderr into logs — previously `cmd.Run()` swallowed all output so we had no idea if the auth even succeeded

PR #1090, opened 2026-04-15
Run Trivy scan on schedule instead of pull requests

torredil:
**What type of PR is this?**

/kind cleanup

The Trivy vulnerability scanner GitHub Action currently runs on every pull request, which blocks and delays dev work. This was discussed with sig-storage lead @xing-yang and we agreed to move it to a daily scheduled run instead. This way we still catch vulnerabilities but PRs to master aren't held up on dep bumps.

Fixes # NA

This is part of a coordinated change across all kubernetes-csi repos that have the Trivy workflow. Same change is being applied across the board.

```release-note
NONE
```

cc: @jsafrane

### DIFF
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -3,7 +3,8 @@ on:
   push:
     branches:
       - master
-  pull_request:
+  schedule:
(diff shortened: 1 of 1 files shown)


andyzhangx (approved):
/lgtm

Issue #927, opened 2025-06-14
Pods with fsGroup can't recover after NFS server restart


vanchaxy:
**What happened**:
After the NFS server restarts, pods can no longer access the mounted filesystem. When the pods are restarted, they fail to start with the following error:

> MountVolume.SetUp failed for volume "pvc-a8667b20-fb7f-4873-aa9d-18ec6f7fdccc" : applyFSGroup failed for vol 192.168.50.210#mnt/user/nfs-csi#pvc-a8667b20-fb7f-4873-aa9d-18ec6f7fdccc##: lstat /var/lib/kubelet/pods/9d3ce1f4-d0c8-4b5e-9994-a17e3267c57d/volumes/kubernetes.io~csi/pvc-a8667b20-fb7f-4873-aa9d-18ec6f7fdccc/mount: stale NFS file handle

**What you expected to happen**:
I expect the CSI driver to detect a stale NFS file handle and remount the NFS share, allowing the pod to recover normally when using fsGroup.

**How to reproduce it**:
1. Create a pod with an fsGroup and fsGroupChangePolicy: "OnRootMismatch", mounting NFS storage provisioned via the CSI driver.
2. Restart the NFS server.
3. Observe that the pod's volume becomes inaccessible (expected).
4. Manually delete/restart the pod.
5. Observe that the pod fails to start with the above error (applyFSGroup failed with stale NFS file handle).

**Anything else we need to know?**:
The CSI driver appears unable to handle stale NFS file handles specifically during the applyFSGroup() operation.

If the pod is restarted without an fsGroup, it starts successfully and the volume is remounted. After this, other pods with fsGroup on the same node also start working again. It seems the stale mount issue is resolved by a remount, but only triggered when fsGroup logic is skipped.

**Environment**:
- CSI Driver version: v4.11.0
- Kubernetes version: v1.33
- OS: Talos v1.10
- Kernel: (can't check at the moment, default from Talos v1.10 — can update if relevant)
- Install tools: Helm chart

cccsss01:
I think I'm experiencing the same problem.

andyzhangx:
it actually fails in the kubelet, related code: https://github.com/kubernetes/kubernetes/blob/7f4f1c29a97de6712d341bf3d724fb122c828772/pkg/volume/csi/csi_mounter.go#L342, not sure what we can do in this csi driver

KodingDev:
Any updates on this?

[...]

PR #1108, opened 2026-04-19
fix: detect and remount stale NFS mounts in NodePublishVolume

andyzhangx:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it:**
When an NFS server restarts, existing mounts on nodes become stale. If a pod with `fsGroup` is then restarted, kubelet's `applyFSGroup` calls `lstat` on the mount path and fails with `ESTALE` (stale NFS file handle) before the CSI driver gets a chance to remount.

This PR fixes `NodePublishVolume` to detect stale mounts and automatically unmount + remount them, so that kubelet's subsequent `applyFSGroup` succeeds.

**How does it work:**
1. When the target path is already mounted, do an `os.Lstat` check
2. If `ESTALE` is detected, unmount the stale mount
3. Fall through to the normal mount path for a fresh mount

**Which issue(s) this PR fixes:**
Ref https://github.com/kubernetes-csi/csi-driver-nfs/issues/927

**Does this PR introduce a user-facing change?**
```release-note
Fix stale NFS mount detection in NodePublishVolume: pods with fsGroup can now recover after NFS server restart without manual intervention.
```

### DIFF
--- a/pkg/nfs/nodeserver.go
+++ b/pkg/nfs/nodeserver.go
@@ -18,10 +18,12 @@ package nfs
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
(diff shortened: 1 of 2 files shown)


andyzhangx:
Addressed the review comment — added unit tests for the stale mount detection path in 3ce23a8:

- Introduced `lstatFunc` package-level variable (defaults to `os.Lstat`) so tests can inject errors
- Added `TestNodePublishVolume` case that injects `syscall.ESTALE`, verifies unmount+remount succeeds
- Added `TestIsStaleFileHandle` covering ESTALE errno, wrapped ESTALE, string-match variants, and negative cases

PR #1115, opened 2026-04-30
doc: document SELinux virt_use_nfs requirement on RHEL-family Enforcing nodes

anvme:
**What type of PR is this?**

/kind documentation

Adds a `case#3` section to `docs/csi-debug.md` documenting that on RHEL-family distributions with SELinux in `Enforcing` mode (default on AlmaLinux, RHEL, CentOS, Fedora, Rocky), the CSI driver's `CreateVolume` and `NodePublishVolume` calls fail with `mount.nfs: Operation not permitted` unless the `virt_use_nfs` SELinux boolean is enabled on every node:

```bash
sudo setsebool -P virt_use_nfs on
```

The error is identical in shape to a server-side export ACL denial, the controller pod has `privileged: true` and `CAP_SYS_ADMIN`, and SELinux denials in privileged containers don't always emit AVC log lines — so this is hard to diagnose without already knowing the answer.

The authoritative gating policy is in [`containers/container-selinux`](https://github.com/containers/container-selinux/blob/main/container.te):

```
tunable_policy(`virt_use_nfs',`
    fs_mount_nfs(container_domain)
    fs_unmount_nfs(container_domain)
    fs_manage_nfs_dirs(container_domain)
    ...
')
```

A single docs paragraph in `csi-debug.md` should help the next operator hit this faster. Tested on AlmaLinux 10.1, kernel 6.12.0-124.49.1.el10_1.x86_64, container-selinux 2.234.2-4.el10, csi-driver-nfs v4.13.2, k3s v1.34.6+k3s1.

Fixes #

Docs-only change, no code or test impact.

```release-note
none
```

### DIFF
--- a/docs/csi-debug.md
+++ b/docs/csi-debug.md
@@ -41,3 +41,54 @@ kubectl exec -it csi-nfs-node-cvgbss -n kube-system -c nfs -- mount | grep nfs
 mkdir /tmp/test
 mount -v -t nfs -o ... nfs-server:/path /tmp/test
 ```
+
+### case#3: `mount.nfs: Operation not permitted` on RHEL/AlmaLinux/CentOS/Fedora/Rocky with SELinux Enforcing
+
+#### Symptom
+`CreateVolume` (controller) and/or `NodePublishVolume` (node) fail with logs like:
+
+```
+mount failed: exit status 32
+Mounting command: mount
+Mounting arguments: -t nfs -o nfsvers=4.1,... <server>:/<share> /tmp/pvc-...
+Output: mount.nfs: Operation not permitted
+```
+
+This happens **even though** the controller and node containers are configured with `privileged: true` and `capabilities.add: [SYS_ADMIN]`. The error is `EPERM` returned by the kernel `mount(2)` syscall — *not* a server-side ACL denial (which would surface as `access denied by server`).
+
+`ausearch -m AVC -ts recent` may show no audit lines, because SELinux denials inside privileged containers are not always logged.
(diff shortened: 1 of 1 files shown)


andyzhangx:
hi @anvme thanks for the contribution, can you sign the easycla first? thx

anvme:
Closing — my diagnosis was wrong.

I posted this PR claiming \`virt_use_nfs\` SELinux boolean is the fix for
\`mount.nfs: Operation not permitted\` on RHEL-family Enforcing nodes. After
further testing on my cluster I traced the actual root cause to a different
issue entirely:

The chart's default mountOptions include \`noresvport\`, which makes the
client connect from a non-privileged source port. On a default
\`/etc/exports\` (which is implicitly \`secure\`), the kernel \`nfsd\` rejects
non-privileged-port requests with \`nfserr_perm\` — surfaced as
\`Operation not permitted\`.

The fix is on the server side: add \`insecure\` to the export options.
SELinux/\`virt_use_nfs\` was a red herring; setting Permissive did not change
the failure on my cluster. The actual reproducer is much simpler than what
I described.

I may file a separate, correctly-scoped doc PR later about the
\`noresvport\` ⇄ \`secure\`/\`insecure\` interaction (which is genuinely a
common stumble — anyone who follows generic NFS tutorial defaults for
\`/etc/exports\` and uses the chart's default mountOptions will hit the same
EPERM I did). Pulling this one to avoid spreading the wrong diagnosis.

Apologies for the noise.

PR #1169, opened 2026-06-22
feat: adds revisionHistoryLimit as a configurable parameter

NPastorale:
/kind feature

Adds `revisionHistoryLimit` as a configurable chart value for `csi-nfs-controller`, `csi-nfs-node` and `csi-snapshot-controller`

Some environments may prefer retaining fewer ReplicaSets to avoid unnecessary clutter, while others might require more. Exposing `revisionHistoryLimit` for all the applicable resources as a chart value provides additional flexibility without changing the current default behaviour.
This is a common parameter exposed by many charts and as such, users may expect it to be configurable (myself included).
Given the infinitesimality and low risk of the change, I've taken the liberty of not opening an issue beforehand. Do let me know whether you'd like me to open one regardless.

- No breaking changes
- Default value remains 10, matching the Kubernetes default
- Auto-formatter fixed a few lines in `values.yaml`, I'm happy to change them back if requested

```release-note
Add `revisionHistoryLimit` as a configurable chart value for `csi-nfs-controller`, `csi-nfs-node` and `csi-snapshot-controller`
```


charts/latest/csi-driver-nfs/values.yaml
```diff
@@ -53,6 +53,7 @@ kubeletDir: /var/lib/kubelet
 controller:
   name: csi-nfs-controller
   replicas: 1
+  revisionHistoryLimit: 10
```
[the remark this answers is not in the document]
NPastorale: Added the new parameters to `charts/README.md`

andyzhangx (commented):
/ok-to-test

charts/latest/csi-driver-nfs/values.yaml
```diff
@@ -1,34 +1,34 @@
 customLabels: {}
 image:
-    baseRepo: registry.k8s.io
```
andyzhangx, line 3: pls keep the all the following indents so it's `image.baseRepo`

NPastorale: I'm sorry, I do not follow. The resulting structure hasn't been modified, only the indentation has been changed from 4 spaces to 2, making it consistent with the rest to the file.
It was corrected by the auto-formatter and I purposefully left it this way to keep it consistent with all the other stanzas.
I can change it back if you wish

andyzhangx: just revert to the original format since it's not related to this PR, thx

NPastorale: it is done

andyzhangx (approved):
/lgtm

Issue #1175, opened 2026-06-25
permissions are set wrong (IMO) on created PV subdirectories


kai-uwe-rommel:
**What happened**:
The provisioner installed fine and seems to work OK so far e.g. it creates a PV in response to a PVC. However, the subdirectory for a PV is created on the NFS server with permissions 755. Thus the pod using it cannot write to it.

**What you expected to happen**:
I would expect the subdirectory to be created with permissions 777.
I otherwise use this provisioner instead: https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner
And this one creates the subdirectories with permissions 777.

**How to reproduce it**:
Just create a PVC and have a pod use it.

**Anything else we need to know?**:
Could I have installed/configured it wrong?

**Environment**:
- CSI Driver version: 4.13.2
- Kubernetes version (use `kubectl version`): OKD 4.22 with K8s 1.35.3
- OS (e.g. from /etc/os-release): CentOS Stream CoreOS (on which OKD is based)
- Kernel (e.g. `uname -a`): 6.12.0-231.el10.x86_64
- Install tools: OKD on vSphere UPI
- Others:

kai-uwe-rommel:
Ok, I now understood I can use the mountPermissions driver parameter and set it to 777 to achieve this effect.
But is that the right way to solve it? If it were, wouldn't it be documented here as such?

I think I could also change the umask for the NFS server process (which currently seems to be 22)?

andyzhangx:
This is working as designed, though I agree the default behavior can be confusing.

### What happens

The controller creates the subdirectory with `os.MkdirAll(path, 0777)`, but the actual permissions are filtered by the process umask (typically `022`), resulting in `0755`. The driver only does an explicit `chmod` when `mountPermissions > 0`:

```go
// pkg/nfs/controllerserver.go
if err = os.MkdirAll(internalVolumePath, 0777); err != nil { ... }

if mountPermissions > 0 {
    // Reset directory permissions because of umask problems
    chmodIfPermissionMismatch(internalVolumePath, uint32(mountPermissions))
}
```

The default value of `--mount-permissions` is `0` (meaning "do not chmod"), so the umask wins and you get `0755`.

### Solution

Set `mountPermissions` in your StorageClass parameters:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: nfs-csi
provisioner: nfs.csi.k8s.io
parameters:
[...]

PR #1238, opened 2026-08-17
docs: explain when to set mountPermissions

andyzhangx:
Fixes the recurring confusion around `mountPermissions` (see [#1175](https://github.com/kubernetes-csi/csi-driver-nfs/issues/1175) and its comments).

Today the one-liner in `docs/driver-parameters.md` documents the mechanics of `mountPermissions` but not the intent. Users hitting `Permission denied` on non-root pods either don't discover the knob, or discover it and immediately set `0777` without understanding the security implication.

## Changes

Adds a new `### Tips` subsection **When to set `mountPermissions`** that:

1. Explains why the default is `0` (the driver deliberately does not override the NFS server's permission model)
2. Recommends `securityContext.fsGroup` on the pod as the preferred fix (the standard Kubernetes pattern)
3. Suggests relaxing the NFS server umask as option 2 (single-tenant exports)
4. Keeps `mountPermissions: "0777"` as a documented last-resort with an explicit warning against multi-tenant / shared-trust-boundary use

Both parameter tables (StorageClass and static PV) now cross-link to this section so users land on the guidance no matter which flow they read.

Docs-only change, no code touched.

### DIFF
--- a/docs/driver-parameters.md
+++ b/docs/driver-parameters.md
@@ -9,7 +9,7 @@ Name | Meaning | Example Value | Mandatory | Default value
 server | NFS Server address | domain name `nfs-server.default.svc.cluster.local` <br>or IP address `127.0.0.1` | Yes |
 share | NFS share path | `/` | Yes |
 subDir | sub directory under nfs share |  | No | if sub directory does not exist, this driver would create a new one
-mountPermissions | mounted folder permissions. The default is `0`, if set as non-zero, driver will perform `chmod` after mount |  | No |
+mountPermissions | mounted folder permissions. The default is `0`, if set as non-zero, driver will perform `chmod` after mount. See [When to set `mountPermissions`](#when-to-set-mountpermissions) below. |  | No |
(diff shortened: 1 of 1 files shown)


andyzhangx:
Addressed the latest Copilot suggestions in `47b2034`:

1. Qualified the `0755` explanation as **dynamic-provisioning only** and added a sentence for the **static-PV** case (mount root keeps the NFS server's existing mode).
2. Added the **`fsGroupPolicy: File` + non-root_squash** prerequisites for `securityContext.fsGroup`, with a link to `deploy/example/fsgroup/README.md`. Also warned about world-writable `0777` not solving cross-GID isolation.
3. Reworded the umask option to talk about the **CSI controller container** (there is no NFS-server-side umask to relax on a stock export).
4. Added a **cluster-wide default** row under option 3 that documents Helm `driver.mountPermissions` (fed into `--mount-permissions` on both controller and node DaemonSet), so Helm users are not told the StorageClass/PV is the only place to set it.

PR #1243, opened 2026-08-18
Don't log mountOptions

stephenfin:
**What type of PR is this?**

/kind bug

`mountOptions` are stored in a secret for a reason. They shouldn't be logged, since logs are frequently collected via centralised logging backends.

n/a

This is a very minor issue with an obvious fix and we don't do stable branches here so there's noting to backport. As a result, I haven't opened a bug report. Please let me know if you'd like one filed.

```release-note
none
```

### DIFF
--- a/pkg/nfs/utils.go
+++ b/pkg/nfs/utils.go
@@ -322,7 +322,7 @@ func WaitUntilTimeout(timeout time.Duration, execFunc ExecFunc, timeoutFunc Time
 func getVolumeCapabilityFromSecret(volumeID string, secret map[string]string) *csi.VolumeCapability {
 	mountOptions := getMountOptions(secret)
 	if mountOptions != "" {
-		klog.V(2).Infof("found mountOptions(%s) for volume(%s)", mountOptions, volumeID)
+		klog.V(2).Infof("found mountOptions for volume(%s)", volumeID)
 		return &csi.VolumeCapability{
 			AccessType: &csi.VolumeCapability_Mount{
 				Mount: &csi.VolumeCapability_MountVolume{


pkg/nfs/utils.go
```diff
@@ -322,7 +322,7 @@ func WaitUntilTimeout(timeout time.Duration, execFunc ExecFunc, timeoutFunc Time
 func getVolumeCapabilityFromSecret(volumeID string, secret map[string]string) *csi.VolumeCapability {
 	mountOptions := getMountOptions(secret)
 	if mountOptions != "" {
-		klog.V(2).Infof("found mountOptions(%s) for volume(%s)", mountOptions, volumeID)
+		klog.V(2).Infof("found mountOptions for volume(%s)", volumeID)
```
andyzhangx, line 325: if that info in the secret is mountOption, we could allow logging, what's not reason for not logging here?

stephenfin: This is a more broad "don't log secrets" issue, or at least don't log them at `V(2)` (which means it's logged by default if using our own assets). Looking through `nfs(5)` it seems you can't do anything silly like provide passwords through mount options (`sec` is maybe the closest thing to this, but knowing that you're using kerberos is not a huge giveaway). I assume there's a good reason we store these in secrets rather than a config map, after all.

andyzhangx: what's because only secret (not configmap) is supported in `DeleteVolume` API:

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: nfs-csi
provisioner: nfs.csi.k8s.io
parameters:
  server: nfs-server.default.svc.cluster.local
  share: /
  # csi.storage.k8s.io/provisioner-secret is only needed for providing mountOptions in DeleteVolume
  csi.storage.k8s.io/provisioner-secret-name: "mount-options"
  csi.storage.k8s.io/provisioner-secret-namespace: "default"
```

